PR #33, opened 2023-07-10
Visual styling of Provenance records page

### LINKED ISSUE #26: Create a bit more visual separation in the provenance
Dear @devhawk,
Can you please create a little more visual separation between provenance records? It is hard for us to tell which timestamp is associated with which record. I recommend the "green stripe" technique of alternating background color for each record, so that no extra real estate is required.

Also, can you make the area where you "Add a Provenance Record" visually separated in some way?

Image

devhawk:
fixes #26

image

### DIFF
--- a/public/ProvenanceForm.js
+++ b/public/ProvenanceForm.js
@@ -17,7 +17,7 @@ export default {
     components: { TagInput },
     render() {
         return html`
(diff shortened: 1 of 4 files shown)


devhawk:
merging w/o review by @RobertLRead request

PR #35, opened 2023-09-14
GIla updates

devhawk:
* Adds optional name to provenance Record
* Displays oldest name value on device provenance page
* Moves create device form to home page
* adds checkbox to create device form to control whether device key is saved in sample device table
* removes memory Repo


RobertLRead:
I approve of this. I ran the test code and it is what we want; thank you.

PR #36, opened 2023-10-17
Rob batch download

RobertLRead:
This adds two entry points:

1. An entry point to allow any number of keys (records) to be created in a single operation.
2. An entry point which does not modify the database, but allows a download of the QR code .png for a device key (presumably, looping over the N keys created with entry point 1.

### DIFF
--- a/server.ts
+++ b/server.ts
@@ -89,16 +89,12 @@ export async function createFastifyServer(deviceRepo: DeviceRepository, recordRe
     // And will return something like:
     // {"keys":["6bXXFunVNY9gtsY47n1tgQ","2UsUGTNYNFWA4jsdWBFxgf","LyGoHQ7wYr9XG1oWMkgXCx"]}
     // ...which creates 3 records and keys with the name "spudboy".
-    type NumberParam = {
-        n: number,
-        name: string
-    };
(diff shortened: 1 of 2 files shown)


server.ts
```diff
@@ -82,6 +83,32 @@ export async function createFastifyServer(deviceRepo: DeviceRepository, recordRe
...
+    // This is Rob, who is a noob here (and the architect, go figure...)
+    // attempting to produce a batch of keys at once.
+    // This can be tested with:
+    // curl -H "Content-Type: application/x-www-form-urlencoded" -X POST http://127.0.0.1:8000/manykeys/spudboy/3
+    // And will return something like:
+    // {"keys":["6bXXFunVNY9gtsY47n1tgQ","2UsUGTNYNFWA4jsdWBFxgf","LyGoHQ7wYr9XG1oWMkgXCx"]}
+    // ...which creates 3 records and keys with the name "spudboy".
+    type NumberParam = { n: number,
+                         name:   string };
+    server.post<{ Params: NumberParam }>('/manykeys/:name/:n', async (request, reply) => {
```
devhawk, line 95: You should be able to  use GET here if you wish since you're passing all the request data in the URL

RobertLRead (commented):
@devhawk I approach you edits; please merge and push.

PR #44, opened 2023-11-03
make provenance descriptions optional

### LINKED ISSUE #40: Allow upload of a new record without requiring a description
Currently the software requires something to be typed into the description. This is unnecessary if they have enterred a tag or uploaded a photo.

This confusion was discovered while doing a demo in front of 15 people.

judithweng:
making provenance descriptions optional

resolves #40

### DIFF
--- a/public/ProvenanceForm.js
+++ b/public/ProvenanceForm.js
@@ -20,7 +20,7 @@ export default {
 <form method="POST" class="col-md-8 card p-3 text-bg-secondary" enctype="multipart/form-data">
     <legend>Create New Provenance Record</legend>
     <div class="mb-3">
-        <input type="text" class="form-control" name="description" id="provenance-description" required placeholder="Provenance Description" />
+        <input type="text" class="form-control" name="description" id="provenance-description" placeholder="Provenance Description (optional)" />
         <label for="provenance-tags" class="form-label mt-3" >Tags (will be converted to lower case and duplicates removed)</label>
         <${TagInput} modelValue=${this.tags} name="tags" id="provenance-tags" class="form-control" onUpdateTags=${(value) => { this.tags = value }}/>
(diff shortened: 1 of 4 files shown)


devhawk:
The `createRecord` method of the `ProvenanceRepository` interface requires a description. If the user doesn't provide value for the `description` form field, what are we [passing to `createRecord`](https://github.com/gosqasorg/asset-provenance-tracking/blob/292be0f87297efb458e6b10b947b2fa7d05c64d4/server.ts#L167)?

```ts
export interface ProvenanceRepository {
    createRecord(key: string | Uint8Array, description: string, options?: CreateRecordOptions) : Promise<ProvenanceRecord>;
    getRecords(key: string | Uint8Array): Promise<readonly ProvenanceRecord[]>;
    getAttachment(key: string | Uint8Array, attachmentID: bigint): Promise<ProvenanceAttachment | null>;
}
```

RobertLRead:
This is a point --- my suggestion to @judithweng was probably incorrect.

What I want us to do is to be able to enter ANY of the fields (not necessarily the description) to create a record. In fact, when we were demoing at MakerFaire (which was happening on stage very fast), people would type in tags, but no description, and not understand why they couldn't update. I think we want a system were you can upload a rating, a simple tag, or a thumbs up without entering a description.

devhawk:
I'm fine with making description optional - either by updating all the places where we can take a description and making it optional or by automatically inserting an empty string as the description if it's not specified in the form. Honestly , I don't know HTML processing well enough to know off the top of my head what you get in the POST handler for an optional form field that isn't specified.

judithweng:
As of now, when the user does not type a description, it is undefined. I can make it pass an empty string instead.

RobertLRead:
@judithweng to continue fixing.

RobertLRead:
The logic here is: You do not need a description if you have already enterred a tag or other info.

RobertLRead:
this PR is no longer relevenat, given the work in the more modern branches we have done.

PR #53, opened 2023-11-15
Move existing app into packages/legacy folder

devhawk:
I *think* this repository structure will allow us to a) continue to deploy the existing (aka "legacy") app to Heroku and b) add new folders under `/packages` for the new front and back ends. I haven't tested, but docs strongly imply that we can deploy Nuxt and Azure Function apps from subfolders in a way that Heroku does not support.

Eventually, after we transition to the new front end and shut down the legacy site, we will remove at least the root heroku Procfile. We can keep the legacy code base around for historical purposes at that point.


RobertLRead (approved):
I haven't studied this in detail, but I approve of it in spirit.

PR #54, opened 2023-11-17
Adding new frontend

### LINKED ISSUE #47: Create new App in Nuxt

K42IE:
Adding the new Nuxt repo to the existing codebase. 

fixes #47


RobertLRead (approved):
I've cast my eyes over this and it looks good to the extent that I understand it. I need to more fully understand how we are keeping the front end separate from the backend.

public/ThumbsRating.js
```diff
@@ -0,0 +1,37 @@
+// Allows user to select a thumbs up or thumbs down rating to be stored with their entry.
+import html from './html.js';
```
devhawk, line 2: Why is this file JS instead of TS?

public/ThumbsRating.js
```diff
@@ -0,0 +1,37 @@
...
+
```
devhawk, line 3: Also, should this file be in the `/packages/frontend/public` folder instead of the `/public` folder?

devhawk (changes requested):
requested changes in code comments

public/ThumbsRating.js
devhawk: Maybe a better question is why isn't this a component?

devhawk (approved):

PR #60, opened 2023-11-26
fix: Makes commands one click copyable & easier to follow

darigovresearch:
See diff


RobertLRead (approved):
These are all impovements in the text. Thanks @darigovresearch! I approve this.

Issue #68, opened 2023-12-03
[BUG] Hamburger menu not working on small screens


darigovresearch:
**Describe the bug**
When clicking the menu in the demo in the UI nothing happens

**To Reproduce**
1. Go to https://gosqas.org/
2. Make width of screen small until the menu appears
3. Click menu button
4. Nothing will happen

**Expected behavior**
Dropdown of menus such as home & devices

**Screenshots**
![image](https://github.com/gosqasorg/asset-provenance-tracking/assets/30328618/a665a4ea-28ae-4d2f-8b2e-36d0b2ed0035)

darigovresearch:
The current gosqas site appears to still have it even if it was merged.

What's needed to remove it in the site?

devhawk:
I need to update the site. Will do shortly

darigovresearch:
Can confirm, thanks for the fix, @K42IE & the update in the UI @devhawk

PR #70, opened 2023-12-05
Removed hamburger menu from legacy frontend

K42IE:
Addressed issue #68


K42IE:
Sample Devices was the only idem in the menu. (I can remove it from the nav if desired too). I edited the code so "Sample Devices" would also show on a small screen.


image

image

RobertLRead (approved):
Great work---thanks for jumping on that immediately!

PR #82, opened 2024-01-02
Create SECURITY.md

RobertLRead:


RobertLRead:
@devhawk can you please approve this very simple addition of a formal security policy?

darigovresearch (changes requested):
Would suggest instead of opening an issue that you can suggest to report it privately using the beta feature available in GitHub which you would need to enable

https://docs.github.com/en/code-security/security-advisories/guidance-on-reporting-and-writing-information-about-vulnerabilities/privately-reporting-a-security-vulnerability

darigovresearch:
Thanks for enabling the report a vulnerability feature!

PR #85, opened 2024-01-20
Container system

### LINKED ISSUE #52: Design container system
Design system drafted in [Provenance Tracking doc](https://docs.google.com/document/d/1zEo3x8z_9rtaa8uSj5YuizLrt8Tz5z_DJQiyNo9J6K8/edit#heading=h.xlt9f63m8fm8) under _Proposed Implementation of the Multi-Level System_

judithweng:
fixes #52 

- Users can enter either a child key or a parent key.
- If a child key is entered, users are given direct access to the child's records (link to the records is automatically given). However, if a parent key is entered, users DO NOT have direct access (link to parent records is not automatically given).
- Users can use "recall" tag to recall all of the devices children, grandchildren, and so on. A record is then added to all children (and grandchildren and so on) with the tag "recall".


RobertLRead (approved):
This was carefully done; in fact my port of it is a bit incomplete. I am merging this, even though it is obsolete, because we need to move to the new system.

PR #105, opened 2024-03-13
Tag Coloring

RobertLRead:
Based on @noramoor2007 's work, I've created what looks like pretty nice tag coloring.

It uses Nora's code, but added a lookup-table for "designed tags" which we wish to have intentional (non-random) colors.  It uses Nora's brightness calculation to use either black or white text.

I'm not an expert at Vue or typescript. I put Nora's code in a utility file called colorTag.ts. In "Feed.vue", where the tags are rendered, I compute the color and textclor directly from the tag in Javascript at time of rendering. This requires no change to the component at all. Personally I find this elegant, but a Vue programmer might prefer a different file.

Screen Shot 2024-03-13 at 9 06 58 AM


K42IE (approved):
Looks good :). Tested.

Issue #106, opened 2024-03-14
[BUG] azureFunc.ts has a hard-wired backend url that needs to be configurable...


RobertLRead:
At present, in azureFunc.ts, the basurl is set:

`const baseUrl = 'https://gosqasbe.azurewebsites.net/api';`

This means you can't run a local backend if you want to. To do so, I comment it out like this:

```
//const baseUrl = 'https://gosqasbe.azurewebsites.net/api';
const baseUrl = 'http://localhost:7071/api';
```

So that it can be overriden.  Assigning to @K42IE for now, but she can send it to Harry or me.

It should probably be done with some pardigm like this:

`const accountName = process.env["AZURE_STORAGE_ACCOUNT_NAME"] ?? "blahblah";`

RobertLRead:
This has been addressed by PR [https://github.com/gosqasorg/asset-provenance-tracking/pull/107](https://github.com/gosqasorg/asset-provenance-tracking/pull/107)

Issue #92, opened 2024-02-11
[FEATURE REQUEST] Report time in local time if we can find a time zone.


RobertLRead:
At present, we are reporting time in UTC (GMT). It would be better to report it based on the local timezone. This cannot be a preference for the user because we don't have authenticated users! But, we can PROBABLY do geolocation from the browser in some cases to learn the preferred timezone.  

The time zone should also be printed, so the user can see it (possibly in a small font.)

Issue #75, opened 2023-12-06
[FEATURE REQUEST] Need a way to understand our usage....


RobertLRead:
At present, users may be using our system, and it is opaque to everyone except a developer who interrogates the database.

I do not want us to break anonymity. However, I recommend that we simply create a new "administrative page" which will not be private, but which we need not advertise, which lets us see (for example):

1. How many devices are in the database
2. How many provenance records are in the database
3. How many devices have been added in the last 24 hours and the last 7 days
4. How many provenance records have been added in the last 24 hours and the last 7 days.
I assume this is a minimal effort (I can't remember if our timestamps are outside of our encryption system or not!) and would give @christinacole, Victoria, and myself everything we need to, for example, say to the World Health Organization, "15,000 new records were added last week, so our system is impactful!"

RobertLRead:
The may need to separate a "frontend dashboard" from the "backend dashboard".

RobertLRead:
Prioritize this lower than getting us switched over to the new Azure system.

vjaqua:
@RobertLRead this sounds like it's related to the statistic page that @suramdivyareddy is working on. Are these issues related?

paulErdos:
Just to note here, this one interests me.

PR #107, opened 2024-03-16
Provide ability to run locally and provide statistics.

RobertLRead:
In this PR, I use use the process.env system and Nuxt.config.ts to allow an override of the default of going to our global Azure blob storage. See Issue #106 This is described in the Readme.

This required changing one of our hooks from "created" to "mounted", a subtle difference.

Additionally I partially addressed issue #75. That issue should probably not be closed---I provide an important basis for it by returning a timestamped history of provenance record additions, but issue #75 specifies additional improvements, such as listing the total number of objects, which I think will be good issues for other programmers.

This also reports times according the the timezone on your browser now, which is issue #92.

I think both @devhawk  and @K42IE should review this, but I would NOT delay merging it unless a problem is found.


RobertLRead:
This functionality was merged in #111 .

PR #111, opened 2024-03-19
Bootstrap is now available...

RobertLRead:
I followed the instructions here: https://stackoverflow.com/questions/70099682/integrating-bootsrap-5-in-nuxt-3-and-vite

...and now, it seems, I can use use Boostrap styling on a page or in a components.

I think we should merge this quickly and @K42IE try it on one of the bugs that she has been trying.

Note:


packages/backend/src/functions/httpTrigger.ts
```diff
@@ -130,10 +131,14 @@ async function getProvenance(request: HttpRequest, context: InvocationContext):
...
         const json = new TextDecoder().decode(data);
         const provRecord = JSON.parse(json) as ProvenanceRecord;
+        const deviceID = await calculateDeviceID(deviceKey);
+        provRecord.record.deviceID = deviceID;
         records.push({ ...provRecord, timestamp });
     }
     records.sort((a, b) => b.timestamp - a.timestamp)
-    return { jsonBody: records };
+  return {
+    headers: { "Access-Control-Allow-Origin" : "*"},
```
devhawk, line 140: CORS is configured at the [service level](https://learn.microsoft.com/en-us/azure/azure-functions/functions-how-to-use-azure-function-app-settings?tabs=portal#cors), so we shouldn't need this.

packages/backend/src/functions/httpTrigger.ts
```diff
@@ -130,10 +131,14 @@ async function getProvenance(request: HttpRequest, context: InvocationContext):
         const { data, timestamp } = await decryptBlob(blobClient, deviceKey);
         const json = new TextDecoder().decode(data);
         const provRecord = JSON.parse(json) as ProvenanceRecord;
+        const deviceID = await calculateDeviceID(deviceKey);
```
devhawk, line 134: Why are we calculating the device key again? It's calculated at the top of the function. And why do we need the device id in each provenance record? I could see adding it once at the top level, but not repeating it in every record

packages/backend/src/functions/httpTrigger.ts
```diff
@@ -187,9 +194,63 @@ async function postProvenance(request: HttpRequest, context: InvocationContext):
...
+        includeMetadata: true,
+        includeSnapshots: false,
+        includeTags: false,
+        includeVersions: false,
+        prefix: ''
+    };
+
+    // Build up a JSON return value
+    // NOTE: We seem to have to read the properties of the blob to get the
+    // metadata.  There is a field called "metadata" on the blob itself
```
devhawk, line 221: I'd have to check the docs, but I think system managed metadata is in that field where user specified metadata is retrived via `blobClient.getProperties()`

packages/backend/src/functions/httpTrigger.ts
```diff
@@ -187,9 +194,63 @@ async function postProvenance(request: HttpRequest, context: InvocationContext):
...
+        // from the addition of new provenance, I supoose.
+        const id = findDeviceIdFromName(blob.name);
+        // We could do some sorting in this function, but that is more or less
+        // easily done by whomever is using this. So I think it better to just
+        // return the data in  a fairly raw form, as an array of {timestamp, id} tuples.
+        // Eventually, this function may have to only look back X days or X hours,
+        // but until it gets unwieldy we can return everything.
+        // I think the proper way to test this is to build a test program that
+        // puts 1000s of objects into the database and see where performance becomes a problem.
+        records.push({ gdttimestamp: metadata.gdttimestamp, gdtid: id});
```
devhawk, line 243: we probably don't need `gdt` prefixes here. And I think we should use the full name `deviceID` not just `id` for this record

packages/backend/src/functions/httpTrigger.ts
```diff
@@ -187,9 +194,63 @@ async function postProvenance(request: HttpRequest, context: InvocationContext):
         const provRecord: ProvenanceRecord = { record, attachments };
         const data = new TextEncoder().encode(JSON.stringify(provRecord));
         const recordID = await upload(containerClient, deviceKey, data, "prov", "application/json", timestamp);
-        return { jsonBody: { record: recordID, attachments } };
+      return {
+        headers: { "Access-Control-Allow-Origin" : "*"},
+        jsonBody: { record: recordID, attachments } };
     }
 }
+// blobNames typically look like: 'gosqas/63f4b781c0688d83d40908ff368fefa6a2fa4cd470216fd83b3d7d4c642578c0/prov/1a771caa4b15a45ae97b13d7a336e1e9c9ec1c91c70f1dc8f7749440c0af8114'
```
devhawk, line 202: remove `typically`. Blob storage isn't exactly like a file system, but blob storage names look like file paths. The device ID *must* be the first segment of the path under `gosqas/` for the system to work

packages/backend/src/functions/httpTrigger.ts
RobertLRead: Removing line 140 DEFINITELY causes a CORS failure in this branch:


Screen Shot 2024-03-19 at 11 49 00 AM

packages/backend/src/functions/httpTrigger.ts
RobertLRead: I think you are right. I have removed this.

packages/backend/src/functions/httpTrigger.ts
RobertLRead: This has been changed to "timestamp" and "deviceID" as requested.

packages/backend/src/functions/httpTrigger.ts
RobertLRead: This has been removed as requested.

packages/backend/src/functions/httpTrigger.ts
devhawk: CORS can be configured when hosting locally via the [local.settings.json file](https://stackoverflow.com/a/55875060). I think we should configure this once in the settings file and remove the explicit code in the httpTriggers.ts file

RobertLRead: I have done as you suggested and tested it.

packages/backend/src/functions/httpTrigger.ts
```diff
@@ -187,9 +189,61 @@ async function postProvenance(request: HttpRequest, context: InvocationContext):
...
+// blobNames look like: 'gosqas/63f4b781c0688d83d40908ff368fefa6a2fa4cd470216fd83b3d7d4c642578c0/prov/1a771caa4b15a45ae97b13d7a336e1e9c9ec1c91c70f1dc8f7749440c0af8114'
+// where the id is that last part (before the last slash)
+function findDeviceIdFromName(blobName : string) : string {
+    return blobName.split("/",4)[3];
+}
+
+async function getStatistics(request: HttpRequest, context: InvocationContext): Promise<HttpResponseInit> {
+    let i = 1;
+    // some options for filtering list
+    const listOptions = {
```
devhawk, line 205: Is listOptions used anywhere?

packages/frontend/package.json
```diff
@@ -12,10 +12,13 @@
   "devDependencies": {
     "@nuxt/devtools": "latest",
     "nuxt": "^3.8.1",
+    "sass": "^1.72.0",
     "vue": "^3.3.8",
     "vue-router": "^4.2.5"
   },
   "dependencies": {
+    "bootstrap": "^5.3.3",
+    "bootstrap-vue": "^2.23.1",
```
K42IE, line 21: Bootstrap vue should not be in the package file. You can remove this by uninstalling it via the terminal

packages/frontend/components/KeyList.vue
```diff
@@ -0,0 +1,14 @@
+<template>
+    <div v-for="key in keys">
+     <a :href="`/provenance/${key}`">{{key}}</a>
+    </div>
+</template>
+
+<script>
+import { myDate } from '~/utils/dateUtility';
```
K42IE, line 8: I don't see where myDate is being used. Also look into using the built in object Date(). If you pass in a timestamp, it will output a readable string.

packages/frontend/components/Providence/CreateRecord.vue
```diff
@@ -26,7 +26,7 @@
 </template>
 
 <script lang="ts">
-import { postProvenance } from '~/services/azureFuncs';
+    import { postProvenance,getProvenance } from '~/services/azureFuncs';
```
K42IE, line 29: Formatting: Add a space after the comma.

packages/frontend/pages/device/[deviceKey].vue
```diff
@@ -1,16 +1,85 @@
...
+            isLoading: true,
+            hasReportingKey: false,
+            childKeys: [],
+        }},
+    async mounted() {
+        try {
+            const route = useRoute();
+            const deviceKey = route.params.deviceKey;
+            const response = await getProvenance(deviceKey);
+            // Warning---I am not sure this works if a provenance has been added.
```
Comment (K42IE) [inline packages/frontend/pages/device/[deviceKey].vue:62]: Is this for the container system? What tests can we do to be sure it functional before pushing the code?

packages/backend/src/functions/httpTrigger.ts
```diff
@@ -187,9 +189,61 @@ async function postProvenance(request: HttpRequest, context: InvocationContext):
         const provRecord: ProvenanceRecord = { record, attachments };
         const data = new TextEncoder().encode(JSON.stringify(provRecord));
         const recordID = await upload(containerClient, deviceKey, data, "prov", "application/json", timestamp);
-        return { jsonBody: { record: recordID, attachments } };
+      return {
+        jsonBody: { record: recordID, attachments } };
     }
 }
+// blobNames look like: 'gosqas/63f4b781c0688d83d40908ff368fefa6a2fa4cd470216fd83b3d7d4c642578c0/prov/1a771caa4b15a45ae97b13d7a336e1e9c9ec1c91c70f1dc8f7749440c0af8114'
+// where the id is that last part (before the last slash)
```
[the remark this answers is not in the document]
RobertLRead: This is fixed.

packages/backend/src/functions/httpTrigger.ts
RobertLRead: No, good catch, it was dead code; I have removed it.

packages/frontend/components/Providence/Feed.vue
```diff
@@ -36,51 +36,42 @@ export default {
...
-    beforeDestroy() { 
+    beforeDestroy() {
         EventBus.off('feedRefresh', this.refreshPage);
     },
     methods: {
         async fetchAttachmentsForReport(report, index) {
             try {
                 if (report.attachments.length > 0) {
-                    const attachmentPromises = report.attachments.map(attachmentID => getAttachment(this.deviceKey, attachmentID));
+                    const baseUrl = useRuntimeConfig().public.baseUrl;
```
K42IE, line 59: Because this is now pulled from the getAttachment function, we should have more documentation in azureFuncs.ts file oh what you need to input into the function.

K42IE: What the baseUrl is not as self-explanatory as deviceKey etc.

packages/frontend/package.json
RobertLRead: Good catch. I have fixed this.

packages/frontend/pages/device/[deviceKey].vue
Comment (RobertLRead) [reply packages/frontend/pages/device/[deviceKey].vue]: I tested this; the comment was old, I should have removed it originally.

packages/frontend/components/Providence/CreateRecord.vue
RobertLRead: Fixed.

K42IE (approved):

Issue #99, opened 2024-03-06
click on image in new front end should open a larger version of the image


devhawk:

RobertLRead:
@K42IE can you test and close this? I think you have already fixed it, can you confirm?

PR #118, opened 2024-03-21
Create Enlarged Image Model onClick with bootstrap

K42IE:


K42IE:
Also fixed bug with date

K42IE:
solved issue #99

RobertLRead (approved):
I think if the Date call on line 24 uses the "myDate" field, we can get automatic time zone internationalization.

PR #130, opened 2024-04-08
Forbidden Tags

K42IE:
This is an addition of Nora's code.


K42IE:
fixes issue #130

RobertLRead (approved):
I approve this. However, I thought the research of finding the forbidden list would be included in this (though it is a somewhat odious and offensive task to find it.) I will create an issue for that.

PR #140, opened 2024-04-28
Fix reporting keys, only admin keys can issue recalls

### LINKED ISSUE #136: [BUG] Fix reporting keys
Reporting keys should not be able to establish parent-children relationships. The current container code allows reporting keys to establish such relationships. This needs to be fixed so that ONLY the admin keys can do such action.

judithweng:
Fixes #136 

Ensures that only admin keys can issure recalls. Any recalls created from a reporting key will not appear in the children records.


RobertLRead (approved):
I tested this by hand. It worked well. @judithweng did a good job.

PR #142, opened 2024-05-13
create form to customize children names

### LINKED ISSUE #98: [FEATURE REQUEST] Ability to name contained devices?
A new form was created for users to be able to create a key for a container and keys for devices within that container, all at once. Currently, users can name the container, add a description, attach an image, create a reporting key and enter the number of devices within the container. However, users are not able to name these devices within the container. As of now, they only follow the name of the container. For example, if 10 masks are inside a container named "box", then the name of each of the masks will be "box # 1", "box # 2" ... "box # 10". 

We need to explore whether this creates a confusion and if users would like to have the ability to name devices within a container.

judithweng:
fixes #98 

Adding input fields for users to customize children names if they want. Children devices are then named after what the user entered.

### DIFF
--- a/packages/frontend/components/CreateContainer.vue
+++ b/packages/frontend/components/CreateContainer.vue
@@ -8,11 +8,24 @@
             <input type="file" class="form-control" accept="image/*" @change="onFileChange" capture="environment" multiple />
 
             <span style="display: inline">
-            <label for="checkbox">Create Reporting Key:</label>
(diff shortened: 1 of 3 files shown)


judithweng:
recent push also fixes issue #143 . Recall function was not able to recall all descendants because of the window reload. Users can now recall all descendants and their window will reload when the form is submitted.

RobertLRead (approved):
I tested the creation of three named devices. It worked well. Great work @judithweng !

PR #147, opened 2024-05-17
Notification Sign up Form

K42IE:
This is currently commented out so it wont show in the UI (because the notification system is not fully functional, I don't want the user to see this yet). This would be a good next step for the design team to tackle. Maybe make a modal etc. 

This is the basic functionality needed. 

image

### DIFF
--- a/packages/frontend/package.json
+++ b/packages/frontend/package.json
@@ -12,7 +12,8 @@
   "devDependencies": {
     "@nuxt/devtools": "latest",
     "nuxt": "^3.8.1",
-    "sass": "^1.72.0",
+    "sass": "^1.77.2",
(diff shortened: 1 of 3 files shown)


packages/frontend/package.json
K42IE, line 1: I needed this via `npm install --save-dev sass sass-loader` for the bootstrap to work properly. Let me know if this needs to be removed etc.

RobertLRead (approved):
This is very nicely done. I approve it. Normally I download all changes and do a light "smoke test" of them when I do a code review. In this case, I don't think there is much point in that---it looks like I would have to uncomment the code to test it. The GUI may or may not be styled attractively, but since I consider that a separate operation anyway, I'm going to accept this.

PR #149, opened 2024-05-24
Match front page and device key page to style guide

judithweng:
Followed style guide created on Figma.

Still need to make minor corrections:
- Toggle switch on the main page, where users choose to make a device or container, does not match style guide but works fine.
- Download QR Code button does not work yet. The download button below the QR code works but I cannot find a way to custom style it.
- Still need to put QR codes for children and reporting keys.


RobertLRead:
You are right -- I tested with better WiFi and the problem went away. We may have to look into that separately.

Nonetheless, I am going to merge this---although I have not really checked it against the style guide I am sure that we are making progress.

RobertLRead (approved):
Dear @judithweng --- this looks like a good use of Bootstrap to me so far. I don't know why I am having bugs on my machine---it could be on my side. (After our conversation, it seems to work.)

Issue #151, opened 2024-05-28
[BUG] Circular Relationships


judithweng:
Currently, there is nothing preventing a user to make a child key be added as a container for the parent / grandparent.

There should be a way to check if the parent that the user wants to add is NOT a descendant of the device.

![circular_relationship](https://github.com/gosqasorg/asset-provenance-tracking/assets/83525666/0dc04e58-c6d0-4814-834b-63f6a0f0f7e9)

PR #153, opened 2024-05-28
preventing circular relationships

judithweng:
fixes issue #151 

This prevents a device from becoming simultaneously both a parent and a descendant of another device. 
circular_relationship

When the user adds a device that creates such relationship, an error is displayed.
adding_child_circular
adding_container_circular

### DIFF
--- a/packages/frontend/components/Provenance/CreateRecord.vue
+++ b/packages/frontend/components/Provenance/CreateRecord.vue
@@ -178,9 +178,17 @@ export default {
                 }
             }
 
-            if (this.childrenKey.length > 1) {
-                console.log("children keys have been entered");
-            } else {console.log("empty children");}
+            if (this.childrenKey.length > 1) { // if children keys have been entered
+                let string_children = this.childrenKey.toString()
+                let list_children = string_children.split(",");
+                for (let i of list_children) {
+                    // for each key, check its descendants and see if current device is a child of them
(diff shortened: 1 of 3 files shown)


RobertLRead (changes requested):
Dear @judithweng unfortauntely, I am with my Mom and have not had time to fully study this. But here is my perception.
1) At lines 162 in CreateRecord.vue, (and elsewhere), I'm afriad the code is looking to make sure something is not 'a child" when really ti needs to be asking "is not a desendant at all". I am not sure, but I suspect the current code, even with your new checks, would  prevent A from becoming a child of B if B is a child of A, but would not prevent A from becoming its own great-grand parent. 
2) You have used the word "notify" in the code and I think it should be changed to "inform". you have made it possible to "inform" all descendants in a recall-style operation.

I did not have a chance to download and test your code before I had to leave to deal with my mom, but can you please try to fix those things?

judithweng:
Hi @RobertLRead, thank you for the feedback!

1. That was probably just a confusing comment but the code does check for all the descendants. In line 163, you can see the variable "descendant list" and it contains a list of all the children, grandchildren, and so on. I will change the comment so it prevents any confusion.
2. Yes, I can change the term from "notify_all" to "inform_all".

RobertLRead (approved):
@judithweng I misread the code, but it is clearer now---thank you for fixing it.

PR #154, opened 2024-05-28
Deploy Front End 

devhawk:
This PR adds a github workflow to auto-deploy main branch to azure Static Web Apps (SWA). The actual step to push the compile result to azure is commented out. Until we create our first GH workflow, there's no way to test this. I think we're better off merging this PR with the deploy step commented out, validate the GH workflow is running correctly, then enable the deploy step.

As part of this work, I had to remove the `packages/frontend` folder from the top level workspaces. Similar reason to why we removed the `packages/backend` folder (https://github.com/gosqasorg/asset-provenance-tracking/pull/109) from the top level workspace - deployment tools appear to depend on a local `node_modules` folder (with workspaces, there is a single shared `node_modules` folder off the root)


RobertLRead (approved):
@devhawk I have looked over all of this. I only partially understand it and I don't know how to test it. I haven't attempted to run it. However, I see no objection, and I don't like to hold up branches, so I think you should merge it. If you insist I try to test it before approving it, I suggest you get @K42IE to do it, since I am out of pocket for some time.

devhawk:
I pinged @K42IE on discord to take a look

K42IE:
My tests were successful after I installed npm in the frontend folder. 👍

K42IE (approved):

Issue #155, opened 2024-06-03
[BUG] Adding children that do not exist or are duplicates


judithweng:
As of now, users can add the same device as children as many times as they want or add keys that do not exist.

The program should not allow users to add containers or children that do not exist. Also, if a device is already an existing child/container, the program should not allow this device to be added again as a child/container.

PR #162, opened 2024-06-14
create error when adding a child that does not exist

judithweng:
fixes issue #155 

When adding a child key that does not exist, the record itself will show "Error: Entered child key does not exist". 

**NOTE** should discuss whether we want to change the language and use different words to convey error message
image


judithweng:
@K42IE would you be able to review this PR? It prevents users from adding keys that do not exist and deletes any duplicates entered by the user.

K42IE (approved):
Tested. Looks good.

judithweng:
@devhawk Resolved all conflicts and ready to merge!

PR #169, opened 2024-06-17
Title bar is now a default layout (on all pages)

K42IE:
fixes #100 

Other fixes:
- when you click the logo it now sends you to the default landing page (need about and contact pages)
- Fixed some design/formatting in the providence page. 

If you have questions about layouts please check out this video: https://www.youtube.com/watch?v=LZDQhOaBBbk

PR #174, opened 2024-06-17
Fix warnings

K42IE:
I made the template into divs. Template should only be declared once in Nuxt3.

This image shows that the fuctionality still works with this method: 
image


devhawk:
looks good, but holding off approval until #169 is merged

devhawk (approved):

PR #188, opened 2024-06-24
Coco update toggle

chihlinc:
- change the toggle design.


RobertLRead:
There is a git-related problem with PR: https://github.com/gosqasorg/asset-provenance-tracking/pull/188 . I believed you merged in a lot of recent (and valid) work, but the upshot is the  PR changes 188 files!  I don't believe this is the right way to do it. I assume to actually fix the toggle (which is the pupose of that PR) there are relatively few lines of code changed. Can you start over (development has more merged) and add into you PR only those lines that effect the toggle?  We can meet about this and do it together if we have to (I know the git issue can be rather subtle.  But this PR is slightly dangerous as it exists right now---I am going to close it. @chihlinc

PR #196, opened 2024-07-07
Working on download function with the file name

chihlinc:


devhawk (approved):
Looks good! Works as expected. Going to hold of on merging until 2024-07-08 meeting since it requires updating the backend which is not automated (yet!)

packages/frontend/services/azureFuncs.ts
```diff
@@ -19,22 +19,34 @@ export async function getProvenance(deviceKey: string) {
...
-        throw error; // re-throw the error if you want to handle it further up the call stack
-      }
+    const response = await fetch(`${baseUrl}/attachment/${deviceKey}/${attachmentID}`, {
+      method: "GET",
+    });
+
+    const blob = await response.blob();
+
+    // Fetch the attachment name
+    const nameResponse = await fetch(`${baseUrl}/attachment/${deviceKey}/${attachmentID}/name`, {
```
devhawk, line 29: Should we check the response headers for `attachment-name` before calling into the `attachment/${deviceKey}/${attachmentID}/name` function? If deploying to the cloud resolves the CORS issue, we can avoid making two round trips every time.

devhawk:
Discussed this PR in today's meeting. Will merge after @chihlinc gets a chance to address my comment above and I have time to manually update the back end after merging

Issue #152, opened 2024-05-28
[FEATURE REQUEST] CSV with children links


judithweng:
Requested by Adam Popanda.

Add the ability to download a .csv file with all of the children links

judithweng:
Fixed thru PR #198 but needs improved format which has been mentioned in issue #210

Issue #190, opened 2024-06-25
Add "Notify All" checkbox in "Create New Record" form


judithweng:
The function Notify All currently exists and works when users add a tag "notify_all". After speaking with Nina, it was concluded that a checkbox will be more intuitive for users. This checkbox is yet to be added to the form. However, the checkbox exists and functions when users create a new container in the home page (see image below).

<img width="827" alt="image" src="https://github.com/gosqasorg/asset-provenance-tracking/assets/83525666/f235b50d-cb21-46f7-93bf-ab2e17ce0653">

judithweng:
Fixed with PR #198

PR #198, opened 2024-07-11
Button to download children keys

judithweng:
This fixes issue #152 and #190.

A checkbox to "Notify All" children has been added to the provenance page. In addition, a button that downloads a .csv file containing children keys and the reporting key has been made. This button can be found at the bottom of the device and provenance page. See image below to view button in device page.
csvbutton

Informed Nina about this new feature and she will be adding this to the figma style guide when she is back from vacation.


RobertLRead (approved):
I've lighted test this, and it looks good, so I am going to approve it. However, the .csv file is being produced like this:

SYc6NjvXg3Qobfwg3Bguhx,SfU4YN5VobWTC8GNhfLRU9

Which I don't think is right. I think we should have each child on a separate line. Possibly the .csv file should have a header as well. I will create a new issue for it.

Issue #189, opened 2024-06-25
Update font sizes, line heights, and spacing to match Figma design


judithweng:
Front page needs some styling fixes. Below is a screenshot of the required font sizes and line heights

<img width="489" alt="image" src="https://github.com/gosqasorg/asset-provenance-tracking/assets/83525666/54631138-d702-4b22-a430-d7628c2f8651">

chihlinc:
@judithweng I can write the codes like the font size below. 
Frontpage - 
Global Distributed Tracking: Heading 1
About, Sample devices, contact: Heading 2
New device, new container: Heading 3
Create new device, device image(optional), Device name, device description, choose files: Heading 4

The Global Open Source Quality Assurance System (GOSQAS) provides the opportunity to create truthful, transparent documentation via our open source provenance tracker, Global Distributed Tracking (GDT). Designed specifically for low-resource settings, GDT allows anyone with simple website access to securely record their project data. Our development philosophy of “Trust Through Transparency” (TTT) encourages better global communication within diverse communities- to include humanitarian response, open source hardware, and scientific research: Heading 5

judithweng:
@chihlinc That's perfect! Thank you!

judithweng:
@chihlinc Katie made some recent changes in PR #194. You can continue to fix the font on the other pages but most importantly try to fix the spacing for the front page. Nina provides the exact padding for each component in [Figma](https://www.figma.com/design/nFA3vMmmL6DdoOIpBmLgIC/Brand-Guidelines?node-id=0-1&t=oJbNwhd4TpSnI0Mp-0).

chihlinc:
What spacing are you referring to? Could you elaborate?

judithweng:
@chihlinc Spacing between each text, buttons and fields. For example, the button "Create Device" currently has a gap between the button and the end of the blue form. This gap looks bigger than the one pictured in Figma. See images below (tops is current site and bottom is figma)

If you could double check if the padding is correct and code it in (if not already), that would be perfect.

<img width="959" alt="gdt-local" src="https://github.com/gosqasorg/asset-provenance-tracking/assets/83525666/8d71bb0a-8f79-4edb-b16c-0d9bc965ea9e">
<img width="515" alt="gdt-figma" src="https://github.com/gosqasorg/asset-provenance-tracking/assets/83525666/a2e8db08-1e3a-4ad2-a3e8-aa1f04255521">

PR #207, opened 2024-07-15
Fixes for issue #189

agilpatr:
Fixes for styling issue #189. Changed font sizes and lines heights to match Judith and Coco's specifications for the front page, changed spacing on the front page to match the FIGMA.


RobertLRead (approved):
I have run this on my local machine, and judging from the files changed there is no risk. I am not checking that it matches the Figma appearance; but I assume @agilpatr did that reasonably well. If something needs to be adjusted, we can do it as a separate issue after this is commited.  I am a bit uncertain about the style of moving from "div" to "h4" for example.  Not that this is wrong, I am just not sure what the style is with Nuxt/Vue/Bootstrap. I am going to merge this, and then send an email to @K42IE about it.

Issue #186, opened 2024-06-21
[BUG] Color tag does not work sometimes


judithweng:
@noramoor2007 I realized that sometimes when I add tags I get the following error (see image below)
<img width="949" alt="colortagerror" src="https://github.com/gosqasorg/asset-provenance-tracking/assets/83525666/ae7952f1-4854-4e6b-ab55-bcb157d9c90b">

I added this random Lorem Ipsum sentence "Bibendum at varius vel pharetra vel turpis nunc eget" and it failed. I did some more testing and found out specifically the word "at" creates this error. I have gotten the error a few times before but I do not remember what words I have used. Would you be able to check if you get the same error and if so, what is creating this issue? Thanks!

RobertLRead:
@judithweng this is an important bug, and I am too busy to do it---can you prioritize this?  I think @noramoor2007 is also too busy to do it---but if she gets to it first, that is fine.

vjaqua:
@judithweng Rob thinks he has merged this, can you verify and close if it has been fixed?

judithweng:
@vjaqua This issue has been fixed. I will close this and mark as solved.

PR #208, opened 2024-07-16
Fixing color tag bug

judithweng:
This fixes issue #186 

The issue seemed to be that the tag "at" was detected as a designed tag but since it is actually NOT a designed tag, it returns an undefined value. I changed the syntax of how we defined `designedTags` and it fixed the issue. Image below shows old code as commented.
designedtagscode

Image below shows tag "at" working. 
tags


RobertLRead (approved):
I tested this. It fixed the bug. I don't quite understand how this code fixes it, but since I tested it I'm going to approve it.

Issue #205, opened 2024-07-15
Create "Jump to Section" And Format the two columns of the History Record Page


K42IE:
Also update the formatting and style of the description, device key, etc. Don't make edits to the feed (this is assigned in this issue: https://github.com/gosqasorg/asset-provenance-tracking/issues/206)
<img width="478" alt="image" src="https://github.com/user-attachments/assets/dd888610-5198-4f87-9e86-53d32484d155">

PR #213, opened 2024-07-19
"Jump To" scroll section in provenance page

judithweng:
Fixes issue #205


judithweng:
This PR contains the 'Jump To' navigation bar on the left side of the provenance page as well as the dropdown version for mobile users. Currently, when you scroll all the way to the bottom of the page of a device with no children, you will notice that the 'Jump To' nav bar won't have the 'Children Keys' section highlighted (see image below). This is not an error but I am hoping that having a footer might solve this issue. If it is still an issue then, I can look into it.
image

RobertLRead (approved):
I love this---it looks good on my desktop. I am concerned about how this would look on a phone, but I will leave that to the designers.

PR #216, opened 2024-07-20
This PR takes an "about" page and "contact" page from Nora's code...

RobertLRead:
I have taken the text from @ninalahoti's Figma. However, the buttons, link and styling need to be improved and brought into conformity. I will assign this to @K42IE to merge.

I am making this PR mainly because @noramoor2007 did some good work that I don't want to lose. At present we don't have a "contact" page in the Figma. Also, we don't have an email function working.

The Figma page calls for a number of buttons (which are in fact simple "links", not form submissions.) However, I am going to leave that for someone else to implement. I think when this PR is merged, an issue should be created for the about page to:
1. Style the page properly (@agilpatr has done this well in the past, I think.)
2. Make the buttons link properly --- not sure what our current style is.
3. Attach the main navigation bar to it.

The contact page is worse in the sense that I don't know how to send the email, and we don't have a design from @ninalahoti for it.

### DIFF
--- a/packages/frontend/layouts/default.vue
+++ b/packages/frontend/layouts/default.vue
@@ -15,13 +15,13 @@
                 <div class="collapse navbar-collapse justify-content-end align-center" id="main-nav">
                     <ul class="navbar-nav ms-2" >
                         <li class="nav-item pe-4">
-                            <a class="nav-link" href="#">About</a>
+                            <a class="nav-link" href="/about">About</a>
                         </li>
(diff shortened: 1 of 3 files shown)


judithweng (approved):
@RobertLRead I changed two lines for the site to link the about page and the contact page so I can see if they work. They look good on my end but we might need to create another git issue to update their styling to match Nina's latest design.

Issue #206, opened 2024-07-15
Format Feed.vue to Match Figma


K42IE:
<img width="557" alt="image" src="https://github.com/user-attachments/assets/4c05965a-2cb7-4c5b-902a-3d8b20986945">

(Only the feed component. Judith is formatting the other parts of the page) 

Figma: https://www.figma.com/design/nFA3vMmmL6DdoOIpBmLgIC/Brand-Guidelines?node-id=668-5291&t=alxePP9lfy6HKFtN-0

PR #217, opened 2024-07-21
Update feed.vue to match the design

chihlinc:
Format Feed.vue to Match Figma #206, and make some frontend design changes such as font, padding


RobertLRead (approved):
I briefly tested this ; it looks okay. I am not qualified to answer if it really looks good against the standard, but it doesn't look worse, so I am going to approve it.

Issue #135, opened 2024-04-17
[FEATURE REQUEST] 


RobertLRead:
This is a rather unpleasant issue. I think we should a list of 100-1000 English words that are not medical terms, but which are either considered obscene or are used to refer to illicit activities or drugs. For example, we should forbid the work "shit" from our list, although we must allow the word "feces", because the words "feces" and "fecal" have important medical uses.

I see no reason these should not be embedded in the code, though we could in theory put them in a separate file; but if we do that, we make it harder for an outside to suggest new forbidden words, such as in Language other than English. For example, I think the word "culo" in Spanish has no medical use; but we cannot hope to head off all obscene words in all languages. By making an open system, we allow other cultures to potentially assist us.

There are other words which we should forbid which are not directly obscene, but have no value in this produce, like "busty". 

There are racial terms that we should exclude.

I do not believe we can forbid a word like "gun" as there may be medical uses of that word.

RobertLRead:
@MairinOGrady I have assigned this to you because it is something a writer could do---it is just the research of a list of words. However, it is a potentially unpleasant and offensive task, if you are squeamish about reading dirty words. I am not, so you can assign this back to me if you would rather not take this on. Or if you want to do some topics but not others, that is understandable.

MairinOGrady:
Hi Robert,

I can start putting together a list. Would this go directly into GitHub or as a Word Document first?

RobertLRead:
@MairinOGrady did a good job with this. I modified her list slightly. I consider it very incomplete, but a first step. 

[Words to Cut Out of Code.v2.docx.zip](https://github.com/gosqasorg/asset-provenance-tracking/files/15266280/Words.to.Cut.Out.of.Code.v2.docx.zip)

judithweng:
Closing this issue as it was solved through PR #218

Issue #80, opened 2023-12-19
[FEATURE REQUEST] Forbid tags which would make illicit use convenient


RobertLRead:
**Is your feature request related to a problem? Please describe.**

We are concerned about future use by bad actors to communicate illicit material

**Describe the solution you'd like**
We want to make it inconvenient for someone to use GDT to transmit illicit material.  We believe making it impossible to add tags related to explicit sexual acts (using non-medical language). 

**Describe alternatives you've considered**
If it is possible to use a service or an algorithm that flag images as pornographic, we would like to use that also. However, we eventually want to allow in-browser encryption, so in that case we won't be able to do that anyway.

**Additional context**
We may want to create other mechanisms which make our system inconvenient for this purpose.

christinacole:
I edited the list to remove legitimate medical words and the names of countries and regions. Banned Words 1.0: https://docs.google.com/document/d/1qQL-Do1mxfN6TbqNOhXu5fG3Ig71QQlMqUY1hmamfAA/edit?usp=sharing

RobertLRead:
@noramoor2007 Can you put these words in place?  I can't remember if you have the system running---but I will assume you can until you assign it to someone else.

devhawk:
fixed by #218

PR #218, opened 2024-07-21
Rob illicit words

RobertLRead:
Fixes issues #135 and #80.

In this code, I use "Rot13" so that we don't have to check naughty illicit words as tags in our code base. Rot13 is a very simple cipher that may only work for the roman alphabet (without diacritical markings.)  We may have to do something else if we need to forbid tags in other languages (UTF-8 character sets, most notably.)

The code uses a "memo" which deciphers the Rot13 strings on the first call. After that, there is no cost. We have about 1100 forbidden words.


judithweng (approved):
@RobertLRead works great! approved!

Issue #201, opened 2024-07-12
About us page


ninalahoti:
Design here: https://www.figma.com/design/nFA3vMmmL6DdoOIpBmLgIC/Brand-Guidelines?node-id=668-5294&t=VlgvyNh92huHZLsA-1

devhawk:
Fixed by #223

PR #223, opened 2024-07-23
Edit CSS from 'About' and 'Contact' page

### LINKED ISSUE #222: [BUG] css of "About" and "Contact" page create issues
The files 'about.vue' and 'contact.vue' contain css code that might be outdated. Some of the code is affecting the design of other pages. For example, all headers are now centered, which is not what every page should have. See image below.
image

judithweng:
@RobertLRead This was something I should have edited before I merged your branch, but I did not realize how every page's style changed with Nora's code. This PR is just a small change and mainly takes out her h1, h2, p, etc. styling format as they are already defined in 'main.scss'. I did not want the main page to look messy so I decided I should quickly fix it.

This fixes #222 

Note: this does not update the styling of the 'About' or 'Contact' page to match it to Nina's design. Nina already created git issue #201 for this.


RobertLRead (approved):
I think this is a wise action. I don't quite understand the issue here, but is seems well justified.

Issue #210, opened 2024-07-16
[BUG] Download of .csv seems to have an improvable format....


RobertLRead:
@judithweng did a great job allowing the children keys to be dowloaded. However, they are dowloaded in a "horizontal" sense and not a "vertical sense".
Here is an example:

SYc6NjvXg3Qobfwg3Bguhx,SfU4YN5VobWTC8GNhfLRU9

I think these should be separated by an newline, instead of a comma. If you have 100s of keys, it a spreadsheet would not want to make them in a single horizontal row.

PR #225, opened 2024-07-24
Changing csv format

judithweng:
Fixes issue #210. Current csv file separates each key in a new line rather than a coma. In addition, each key has been converted to a link to its provenance. This is the current format: https://gosqas.org/provenance/{key}.


RobertLRead (approved):
I have tested this. The code looks clean and simple.

Issue #170, opened 2024-06-17
Make the heading dynamically size (to fix hamburger menu layout issue 


K42IE:
Make sure that this is in one line by dynamically sizing the logo (or both the logo and hamburger menu) to fit the screen size.

Also, please change the current hamburger menu icon to match the Figma. 

Wanted Result: 
![image](https://github.com/gosqasorg/asset-provenance-tracking/assets/25025541/e1ca36dd-a5ef-4b90-8a23-8c0502df25c6)

Current:
![image](https://github.com/gosqasorg/asset-provenance-tracking/assets/25025541/35a2d2c4-9bc6-4e2b-ad64-79f8ef70f691)

![image](https://github.com/gosqasorg/asset-provenance-tracking/assets/25025541/f78cde90-73c8-4891-a3a8-7312c28ccc6c)

ninalahoti:
<img width="564" alt="Screenshot 2024-06-17 at 1 02 47 PM" src="https://github.com/gosqasorg/asset-provenance-tracking/assets/71244080/70da11b7-8fb3-4b8d-a55d-19e51f774eea">
Can we get it closer to this screenshot, please?

PR #231, opened 2024-07-30
Fix hamburger bar styling

agilpatr:
Fixes for issue #170. Made the icon match the FIGMA and fixed the issue where the hamburger bar would move to a new line when the screen got too small. Screen sizes smaller than 187px will go back to two lines, but to my knowledge no screen sizes are that small so it shouldn't be an issue.


RobertLRead (approved):
I tested this and compared it against our Redstone test site. It seems to solve the problem.

Issue #192, opened 2024-06-27
Download Button on Device Page


K42IE:
Task One: Remove Grey Download Button :![image](https://github.com/gosqasorg/asset-provenance-tracking/assets/25025541/049655a5-0b41-425e-88da-2061eb3c579a)

Task Two: And make sure the styled button actually downloads the QR code (it currently does not).

Task 3: Nina also wants a higher resolution QR code

chihlinc:
Okay!

chihlinc:
PR  #233

PR #233, opened 2024-08-01
New download button and change the QR code design

chihlinc:
1. QR codes at the moment look different but lead to the same link
2. I cannot add a logo in the center. The QR code disappeared whenever I tried to load a logo in the center.


RobertLRead (approved):
This looks good. I think we are making progress. I don't undestand why the QR codes has rounded edges.

RobertLRead:
Fixes #192

PR #238, opened 2024-08-07
Coco change upload format - accept everytype of files, including pdf. 

chihlinc:


RobertLRead (approved):
This appears to work. Good job, @chihlinc !

PR #243, opened 2024-08-18
Created a button component

agilpatr:
Made a button component. It doesn’t work for every type of button, but it should work for the simple ones (ex. purple buttons, outlined buttons, etc). You can specify color, size, and spacing through the button component. The text size is hardcoded to be the specified button size from a previous issue (I didn’t see anywhere that had buttons with different text sizes in the FIGMA) but custom sizes shouldn’t be too hard to add later if needed.

I replaced all of the buttons I could find with the new component, though there were a few I skipped/likely missed. I didn’t change the buttons in [deviceKey].vue (in the provenance folder) since the button layout was very different to the button component I made. I also didn’t change one of the buttons in [deviceKey].vue (in the device folder), as I couldn’t get it to take the user to the correct page using the button component.


packages/frontend/components/ButtonComponent.vue
K42IE, line 1: I recommend calling it "Button" or "CustomButton"

K42IE: Also make sure you "git pull" to keep your code up to date!

judithweng:
@agilpatr Thank you for making the button component! Your PR had a few conflicts but I made one notable edit. I took out your edits from device/deviceKey.vue. This file contains two buttons: "View Provenance Records" and "Download QR Code". For the view records button, you mentioned that you were not able to get the link to work, and for the download QR, Coco made new edits that your branch did not have. Hence, I left these two buttons as they were originally. 

We can make this a separate git issue if we want to look deeper into why the link did not work.

judithweng (approved):
Button works for create device, create container and create new record.

Issue #117, opened 2024-03-21
[FEATURE REQUEST] Automatically convert URLs in prov records to clickable anchor tags


devhawk:

RobertLRead:
@noramoor2007 I think this is a relatively low priority bug, but you might want to do it after you get the word filter working. It is similar in nature, in that it can be a piece of pure javascript that runs in the browser (not the backend). This could be the first issue where you are actually directly coding in the frontend.

RobertLRead:
Note: This issue concerns only making a URL clickable, not anything more complicated. This is usually done with "anchor" tags --- by convention a clickable link is often rendered as an underlined blue text.

PR #247, opened 2024-08-19
Clickable Link Code Addition #117

anushashringi042:
This is the pull request for #117 where I created a clickableLink.ts folder addition that way any links added on provenance records appear as links that can be clicked.


judithweng:
Implemented Anusha's code into our site. The current code allows any link to be clickable but the link must start with http or https. Links with just the domain will not be clickable (see image below). We can change this if it's not desired.

image

RobertLRead (approved):
This is really nice!  I love this.

PR #249, opened 2024-08-19
Issue with Attaching Non-PNG Files to Containers and Records

chihlinc:
Allow every other file type to containers and records


K42IE:
image
it says download image for a pdf and a word doc. This is misleading. Fix or create a new issue to fix @judithweng

K42IE:
image
Otherwise it seems to work for me 👍

judithweng (approved):

PR #253, opened 2024-08-27
Adding buttons to pages

judithweng:
This is a very large PR that essentially styles the 'Home', 'About Us', 'How it Works' and 'Data & Privacy' pages. The specifics are listed below.

1. Fixes #244 : The previous PRs #234 and #226 created the 'How it Works' and 'Home' page. However, these did not contain buttons. In this PR, the buttons are added thanks to Ally's button component #243.
2. Fixes #201 : An 'About' page already existed from one of Nora's old PR. This page had most of the contents but did not have the latest design. In this PR, the 'About' page follows the latest Figma design.
3. Fixes #203 : A 'Data & Privacy' page was made in this PR following the latest Figma design.

Things missing or to fix:
1. The 'Home' page is still not our official home page, but I think once this PR is approved, then we can immediately make this the home page. To visit this 'Home' page, use `http://localhost:3000/home`

2. Sometimes when I click 'go back' on my browser, I am unable to see the page (see image below), but it works fine when  I reload it. This problem, however, ONLY happens in dev mode (with `npm run dev`) but not in production mode (`npm run preview`).
image

Issue #248, opened 2024-08-19
Style "create new history record" section


ninalahoti:
New design here: https://www.figma.com/design/nFA3vMmmL6DdoOIpBmLgIC/Brand-Guidelines?node-id=668-5291&t=IS74iTaJWXdDGuiS-1

PR #256, opened 2024-08-29
'Create New Record' form

judithweng:
This PR solves #248. The 'Create New Record' form has been styled to match the latest Figma design.

As stated in PR #253, the button components have a font size that cannot be changed. The button to submit a record does not have the correct font size and needs to be changed.


RobertLRead (approved):
I have lightly tested this. I do not completely understand the difference of the files, but I have great faith and I think this is worth moving forward.

PR #275, opened 2024-09-18
Remove legacy package.

bencoombs:


RobertLRead:
Glad you made this --- don't merge till we get the migration done. Harry worked on it tonight and ran into a SNAFU but we should get it fixed pretty soon.

RobertLRead (approved):
I think it is time for us to remove this now. I see there are some conflcts (code always "bitrots" in that way.)  If you are careful about the conflicts, I think we can go ahead and merge this. We still have the site running at Heroku but I can't see why we would want to build the legacy package again.  Please merge, or ping @devhawk if you want a second opinion.

PR #278, opened 2024-09-19
Add test framework

bencoombs:
- Added vitest to frontend and backend packages.
- Added test commads in backend and frontend directories.
- Running 'npm test' in the root directory will run the backend and frontend tests with coverage. This can be called from CI in the future.
- Added example tests for the frontend and backend.

image
image


RobertLRead (approved):
This looks like real good, careful, thoughtful work. I've reviewed all the code changes; I have not pulled it down and tested it. I'm pretty busy with some other things so I think this is worth approving without a test, although I normally would test it. I will leave it for @bencoombs to merge at a time convenient for him.

PR #279, opened 2024-09-19
Make 'Device Provenance Record' page match Figma

bencoombs:
- Change device key to device name


bencoombs:
This branch is for Melanie to complete.

PR #282, opened 2024-09-24
upgradeProvenance

devhawk:
* add `convertLegacyProvenance` function that converts legacy prov records/attachments under the `gosqas/legacy` folder
* added `upgradeProvenance` HTTP endpoint that invokes `convertLegacyProvenance`
* added `convertLegacyProvenance` call in `getProvenance` if there are no prov records for a given key.
* extracted prov upload logic from `postProvenance` into `uploadProvenance` so it can also be called from `convertLegacyProvenance`

### DIFF
--- a/packages/backend/package-lock.json
+++ b/packages/backend/package-lock.json
@@ -1,4 +1,3 @@
-
 {
   "name": "gosqas-az-func",
   "version": "0.0.0-placeholder",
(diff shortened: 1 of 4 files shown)


RobertLRead (approved):
I approve this. I have not tested it --- I believe @devhawk has been testing as it goes. The code quality looks good to the naked eye.

PR #284, opened 2024-09-24
Added custom font sizes to ButtonComponent.vue

agilpatr:
I implemented custom text sizes by making a custom css class in ButtonComponent that can be set to either true or false when creating a button (I give an example of this in the comments at the bottom of ButtonComponent). To make a new custom button size you would have to make a new css class, it’s a little messy but it works.

### DIFF
--- a/packages/frontend/components/ButtonComponent.vue
+++ b/packages/frontend/components/ButtonComponent.vue
@@ -1,6 +1,7 @@
 <!--Button Component-->
 <template>
-   <button :class="[example ? 'example' : 'baseButton']" :style="styles">{{ buttonText }}</button>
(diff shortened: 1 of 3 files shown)


bencoombs:
Looks great Alison!

Will there be many custom button sizes?

agilpatr:
Hi Ben,

We can make as many custom button sizes as we need, Judith gave 2 examples in the issue but there might be more. Judith only gave desktop example sizes in the issue (and I'd need desktop/mobile sizes and font-weight to add a new style class), so I only added an example one and the default button sizing.
________________________________

bencoombs (approved):
Awesome work!

Issue #285, opened 2024-09-25
[BUG] Footer doesn't stick to the bottom of the page


bencoombs:
**Describe the bug**
The footer doesnt stick to the bottom of the page if there is not enough content to fill the height of the page.

**To Reproduce**
1. Go to any page
2. Zoom out 
3. Observe that the footer moves up the page

**Expected behavior**
The footer should stay at the bottom of the page.

**Screenshots**

**Desktop (please complete the following information):**
 - OS: Windows
 - Browser: Chrome, Firefox
 - Version: latest

bencoombs:
![image](https://github.com/user-attachments/assets/bb5ec6d7-7659-44bc-98b9-96baa98faf2b)

![image](https://github.com/user-attachments/assets/3e63b0a3-ea8d-4b06-9a7d-fb4bb6cb3588)

bencoombs:
PR: https://github.com/gosqasorg/asset-provenance-tracking/pull/286

PR #286, opened 2024-09-25
Make footer sticky.

bencoombs:
Update the styling to make the footer stick to the bottom of the page if there is not enough content.

Issue: https://github.com/gosqasorg/asset-provenance-tracking/issues/285

### DIFF
--- a/packages/frontend/assets/styles/main.scss
+++ b/packages/frontend/assets/styles/main.scss
@@ -49,12 +49,22 @@ $form-file-button-bg: #4e3681 ;
 $form-file-button-hover-bg: #322253;
 
 
-
-// import bootstrap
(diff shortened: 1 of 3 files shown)


noramoor2007 (commented):
This is good, I think it should be merged.

One issue is that when you zoom out, the main body text stretches to the edges of the screen instead of remaining centered, leaving a large space between the body and the footer. Ideally, the text should stay centered to prevent this excessive space. This is what I mean:
image

Using IBM's platform as a reference, you can see how they have their footer stuck to the bottom of the page, which is what you have already done well here. However, when you zoom out on IBM's page, while body of the webpage shrinks the left and right margins increase proportionally:
image

This is minor, but it should maybe be considered.

noramoor2007 (approved):

PR #287, opened 2024-09-25
Fix page not rendering correctly after creating a new device.

bencoombs:
- Added an await to the router navigation call - this is likely the cause of the bug.
- Added navigation error reporting to gather more data in case the bug is encountered again.


RobertLRead (approved):
I have not tested this, but the reasoning seems clear.

PR #288, opened 2024-09-26
Update QR Code does not work and fix its domain 

chihlinc:


bencoombs (changes requested):

bencoombs (approved):

bencoombs (changes requested):
Sorry I think we need to revisit this PR.

There is another localhost:3001 in GenerateQRCode.

Is the QR code link supposed to go to the website?

bencoombs:
Trying to fix this wasn't as simple as I thought, since the QR download button is not part of the QR code component. I had to add a reference for the parent component (with the download button) to access the the child components data (the QR code).

This QR code is accessed by a singleton code factory object which is setup as a Nuxt plugin. This is from https://dev.to/benuuts/how-to-use-qr-code-styling-in-nuxt-3-1pno

The qr code was also defined twice - once programmatically and once as a component. I have not used the Vue component and consequently we also no longer need the forked QR code library - in a separate PR, we should be able to remove the unused code.

FRONTEND_URL can be set as an env var to set the QR code base url.

bencoombs:
> Sorry I think we need to revisit this PR.
> 
> There is another localhost:3001 in GenerateQRCode.
> 
> Is the QR code link supposed to go to the website?

From this weeks meeting, the link should be the frontend URL that can be configured as a env var, like we have the baseUrl set.

RobertLRead (approved):
This looks real good! Thank you for doing this.  I have one quibble---I have always thought the "roundedness" of our QR codes decreased their effectiveness if printed very small, or if one tried to 3D print them onto a device. Can we remove that and just make them  "square"?  This is more of a hunch of mine than a known fact.

bencoombs:
image
I've made the corners square.

PR #290, opened 2024-10-01
Fixed thumbnail issue.

noramoor2007:
Screenshot 2024-09-30 at 9 48 53 PM

### DIFF
--- a/packages/frontend/components/Provenance/Feed.vue
+++ b/packages/frontend/components/Provenance/Feed.vue
@@ -39,10 +39,6 @@ along with this program.  If not, see <https://www.gnu.org/licenses/>. -->
             
             <div v-for="(attachment, i) in attachmentURLs[index.toString()]" :key="i">
                 <!-- Image -->
-                 <!--THIS WAS THE ISSUE:-->
-                 <!-- <div v-if="((attachment.fileName).toLowerCase()).endsWith('.jpg') || ((attachment.fileName).toLowerCase()).endsWith('.png')  ">
-                    <img :src="attachment.url" :alt="Image" style="width: 150px; padding: 5px;" data-bs-toggle="modal" data-bs-target="#imageModal" @click="modalImage = attachment.url">
-                </div> -->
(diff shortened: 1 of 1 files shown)


packages/frontend/components/Provenance/Feed.vue
```diff
@@ -39,9 +39,7 @@ along with this program.  If not, see <https://www.gnu.org/licenses/>. -->
             
             <div v-for="(attachment, i) in attachmentURLs[index.toString()]" :key="i">
                 <!-- Image -->
-                 <div v-if="((attachment.fileName).toLowerCase()).endsWith('.jpg') || ((attachment.fileName).toLowerCase()).endsWith('.png')  ">
-                    <img :src="attachment.url" :alt="Image" style="width: 150px; padding: 5px;" data-bs-toggle="modal" data-bs-target="#imageModal" @click="modalImage = attachment.url">
```
bencoombs, line 43: Can you please include the full img tag because the end part opens a modal with a larger image. This part should be added back: data-bs-toggle="modal" data-bs-target="#imageModal" @click="modalImage = attachment.url"

packages/frontend/components/Provenance/Feed.vue
```diff
@@ -39,9 +39,7 @@ along with this program.  If not, see <https://www.gnu.org/licenses/>. -->
...
-                 </div>
```
bencoombs, line 44: Can you please include the full img tag because the end part opens a modal with a larger image. This part should be added back: data-bs-toggle="modal" data-bs-target="#imageModal" @click="modalImage = attachment.url"

bencoombs (changes requested):

bencoombs:
> Have you tested this with .jpeg, .jpg, .png, .gif? There are a bunch of others that we should define and write tests againsts, but this should cover the basics.

As discussed as the weekly meeting, this PR will resolve the immediate issues and we will address support for further file types etc. as part of a separate effort.

bencoombs (approved):

PR #298, opened 2024-10-07
Fix scrollspy hidden behind footer

bencoombs:
Changing the body tag to div tag fixes the rendering issues where the scrollspy is hidden behind the footer. There should only be one body tag.

image


RobertLRead (approved):
I don't fully understand this, but it seems well worth merging.

PR #306, opened 2024-10-08
Child keys not rendering correctly

bencoombs:
Child keys are rendering as "undefined". From real data it was observed that the child key can be undefined or [].

The root cause is the search for the child key isn't checking for both cases.

The code is moved into helper functions that can be unit tested against a real example of the provenance data structure (with fake data).


RobertLRead (approved):
To my eye, this looks like an impressive use of Typescirpt/ECMAscript. I think you are doing a better job with the language specifics than I could. It looks to me like you are improving the code quality a good bit!

PR #310, opened 2024-10-11
Fix undefined error.

bencoombs:
Protect against undefined childKeys. Updated device page to be the same as the provenance page - this could be broken out into a component later.

Hotfix for: https://github.com/gosqasorg/asset-provenance-tracking/pull/306


RobertLRead (approved):
I don't fully follow the Nuxt changes --- but I trust they are good. I understand the typescript which handles the nullables more, however.

PR #315, opened 2024-10-14
Coco email sending notification

chihlinc:


packages/backend/src/functions/httpTrigger.ts
```diff
@@ -400,3 +402,8 @@ app.get("getStatistics", {
     route: 'statistics',
     handler: getStatistics
 })
+
+
+
```
bencoombs, line 407: Are you able to create a new endpoint here that we can use to test the email client is working in Azure?

RobertLRead:
@bencoombs  @chihlinc what is the status of this PR?  It seems like it has not been worked on for some time.

bencoombs:
Duplicate branch - email functionality was merged in already.

PR #338, opened 2024-11-18
Fix broken props.

bencoombs:
Previous renaming of device to record broke the add record to feed functionality.


RobertLRead (approved):
I approve this (obviously, it fixed the problem.)  I'm a little surprised by this though---was this introduced by @agilpatr and not caught in testing?  In any case, please merge it ASAP.

bencoombs:
There were 2 issues:
1.  some props being renamed and some weren't which caused them to break.
2.  recursive search of the descendants failing - I've just changed it to get the children since that's all we need for the group right now.

I did test it manually with a local version but obviously not well enough! We have a test framework in place, and are progressing on the devops front, so we should be more robust to these regressions in the near future.

PR #346, opened 2024-11-20
Update logging for backend.

bencoombs:
Set log levels for Function app to Information to get them to show context.log statements.


RobertLRead (approved):
This is interesting. I see that this is a built-in functionality of Azure. I don't quite know how it works or where it goes, but given my limited knowledge of this, I approve it. I am glad you are giving us this new-ish capability.

PR #351, opened 2024-11-23
Remove notify all checkbox on record form

bencoombs:
Remove notify all checkbox on record form


RobertLRead (approved):
I approve this. This is so simple I will merge this myself.

Issue #339, opened 2024-11-19
Firefox issues


bencoombs:
![Image](https://github.com/user-attachments/assets/ecf00e73-daac-44c9-8542-52413cb35cf7)
- Tried to add group key then received error: "Container could not be added" 
- Create Record Entry was not functional
- Browser: Firefox

(there are known issues with Firefox support)

Issue #264, opened 2024-09-11
Create the notification service


bencoombs:
Create the code scaffold for the notification service. A separate service is desired so that the service can be used independently of GDT and to prove the technical requirements of the service are met.

Issue #347, opened 2024-11-22
[FEATURE REQUEST] Loading indicator


bencoombs:
**Is your feature request related to a problem? Please describe.**
On slow internet it can take up to 15s to load the page.

**Describe the solution you'd like**
Showing a loading indicator would make it clear to the user that data is still downloading.

bencoombs:
https://github.com/gosqasorg/asset-provenance-tracking/issues/352

PR #355, opened 2024-11-29
UX improvements

bencoombs:
A few changes to address some bugs and to improve the user experience (based on user feedback).
- Added helper function to validate key.
- Added helper functions to determine if a record is a group.
- Added helper function to search over the list of immutable records for changes, eg. adding child keys or adding a record to a group is appended, so it may not be in the first record. Now can search for 'features' in the record and then use those in the app.
- Improved load times by reducing number of HTTP calls; actual reduction in calls and also increased client-side error checking to reduce superfluous calls.
- Improved user experience by only updating the feed component instead of reloading the entire page.
- Added enumerated tags.
- Added very simple loading text.

### DIFF
--- a/packages/frontend/components/Provenance/CreateRecord.vue
+++ b/packages/frontend/components/Provenance/CreateRecord.vue
@@ -120,9 +120,8 @@ export default {
         async submitRecord() {
             // Get a refreshed copy of the records
             const records = await getProvenance(this.recordKey);
-            console.log("Exusting records", records);
             
-            if (!records) {
+            if (!records || records.length === 0) {
                 this.$snackbar.add({
                     type: 'error',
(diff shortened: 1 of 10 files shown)


bencoombs:
Closes #339 #347 to fix high priority user-reported issues.

Partial progress on #24 by adding tag enums.
Closes #264 sufficiently to provide a foundation for additional notification features.

RobertLRead (approved):
I haven't tested this, but looking over all of the code, it looks really great. It looks like a lot of thoughtful work was put into making this nice well-written code.

bencoombs:
Rebased onto main and tested locally. Added snackbars to bring this in line with the rest of the UX.

Issue #332, opened 2024-11-10
Update site to match Figma


vjaqua:
Update gosqas.org to match Figma https://www.figma.com/design/nFA3vMmmL6DdoOIpBmLgIC/Brand-Guidelines?node-id=0-1&node-type=canvas&t=xapC5Rtayo8Fx46Y-0

1. Add Terms & Conditions page
2. Add Homepage
3. Review Figma for additional updates

ninalahoti:
Will add more notes once these edits are complete.

RobertLRead:
@agilpatr is making progress plans to have done by Dec. 5th.

agilpatr:
2. Add Homepage refers to moving the homepage to https://gosqas.org/ and moving the page currently there to https://gosqas.org/gdt

PR #358, opened 2024-12-07
Changed frontend to match FIGMA

agilpatr:
Implemented the changes to the site specified in issue #332. The only part I didn't fix was the blurry images, since the blurry images are smaller files than the non-blurry images (and since I don't have the original art I can't download them at a higher resolution). I also tried getting SVG versions but it wasn't working for me, it just ended up changing how the images looked when I did it.


packages/.DS_Store
bencoombs, line 1: This shouldn't be commited. Can you please remove this and add the file extension to gitignore?

packages/frontend/layouts/learn_more.vue
```diff
@@ -20,10 +20,19 @@ along with this program.  If not, see <https://www.gnu.org/licenses/>. -->
 <template>
 
     <div class="row" id="learn-more">
-        <h3 class="text-iris"> Interested in learning more? </h3>
-        <p style="font-weight: 400;"> Learn more about Global Open Source Quality Assurance System (GOSQAS) by emailing us at 
-        <span style="font-weight: 400;" class="text-decoration-underline text-iris"> gosqasystem@gmail.com</span> 
+        <h2 class="text-iris"> Interested in learning more? </h2>
+        <p class ="mb-0" style="font-weight: 400;"> Learn more about Global Open Source Quality Assurance System (GOSQAS) by emailing us at 
+        <span style="font-weight: 400; overflow-wrap: break-word;" class="learn-more-desc text-decoration-underline text-iris"> gosqasystem@gmail.com</span>
```
bencoombs, line 25: Can the style attributes be moved to the class learn-more-desc?

packages/frontend/pages/gdt.vue
```diff
@@ -0,0 +1,174 @@
...
+MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+GNU Affero General Public License for more details.
+
+You should have received a copy of the GNU Affero General Public License
+along with this program.  If not, see <https://www.gnu.org/licenses/>. -->
+<!--
+    This is the landing page where you can create a new record to track
+-->
+<template>
+    <link href="https://fonts.google.com/specimen/Poppins" rel="stylesheet" type="text/css">
```
bencoombs, line 24: Does this successfully load for you? I've always got a MIME type warning when getting font but I haven't looked into it.

packages/frontend/pages/index.vue
```diff
@@ -18,57 +13,89 @@ GNU Affero General Public License for more details.
...
+
+        <div class="row bg-frost" id="second-row" >
+            <div class="row rowtest">
+                <div class="col test" id="second-row-cols" v-for="item in second_row">
+                    <h3 class="text-iris" id="second-row-cols-h3">{{item.title}}</h3>
+                    <p class="text-eggplant" style="font-weight: 400;">{{ item.descr }}</p>         
+                </div>
+            </div>
+            <div class="col" style="text-align: center; margin-top: 50px;">
+                <button-component class="button" buttonText="About Us" color="#322253" onclick="window.location.href='about'"
```
bencoombs, line 54: This works but to be Vue-like can you please create a method that uses the vue router to navigate to the page?

packages/frontend/pages/index.vue
```diff
@@ -18,57 +13,89 @@ GNU Affero General Public License for more details.
...
+
 
 <script lang="ts">
+import Learn_more from '~/layouts/learn_more.vue';
+import ButtonComponent from '~/components/ButtonComponent.vue';
+let showTrack = false;
+
+const second_row = [
+    { title: "Simplicity & Accessibility", descr: "We believe that open-source projects should be simple to use and understand."},
+    { title: "Data Ownership", descr: "We do not have access to any user data, ensuring complete privacy and independent ownership."},
```
bencoombs, line 75: Please can you use the full word "description"

packages/frontend/pages/index.vue
```diff
@@ -18,57 +13,89 @@ GNU Affero General Public License for more details.
...
-            if (toggle.checked) {
-                createRecord.style.display = "none";
-                createGroup.style.display = "block";
-            } else {
-                createRecord.style.display = "block";
-                createGroup.style.display = "none";
+        // Function to have the 'Track an asset' input field appear
+        async trackingForm() {
+            let trackAssetDiv = <HTMLDivElement>document.getElementById("homeTrackAssetDiv");
+            let trackButton = <HTMLDivElement>document.getElementById("homeTrackButton");
```
bencoombs, line 84: This works but we should refrain from accessing the DOM directly and try find a way to do this using Vue.

packages/frontend/pages/index.vue
```diff
@@ -77,93 +104,102 @@ export default {
...
-    display: grid;
-    grid-template-columns: repeat(2, 1fr);
-    width: fit-content;
-    border: 2px solid #4a148c;
-    border-radius: 25px;
-    background: white;
-    cursor: pointer;
-    overflow: hidden; /* Ensure no overflow */
+#first-row {
+    background-image: url(../assets/images/hand-icon.png);
```
bencoombs, line 120: Can this go in the nuxt config? That way all the configuration is in one place.

bencoombs (changes requested):
Awesome effort! Only minor changes requested.

packages/frontend/pages/gdt.vue
agilpatr: Not sure what that line was doing as the font style comes from elsewhere, I just removed it as it seems to have no effect.

packages/frontend/pages/index.vue
agilpatr: If you use the word 'description' instead of 'descr' then the text doesn't appear on the screen.

packages/frontend/pages/index.vue
agilpatr: I didn't work on this part originally, so I am struggling to find a way to do this with Vue. For the sake of getting higher priority issues completed I would maybe make a new issue for this.

packages/frontend/pages/index.vue
bencoombs: OK, can you please create the issue?

packages/frontend/pages/index.vue
bencoombs: OK, can you please create the issue?

bencoombs (approved):

packages/frontend/pages/index.vue
agilpatr: Both requested issues have been created.

PR #360, opened 2024-12-08
Run npm audit fix

bencoombs:
Fix 2 high vulnerabilities.
Unfortunately this updates creates lots of harmless deprecation warnings because of a SASS update. We have to wait until Bootstrap releases an update.


bencoombs:
Completed in another PR.

PR #363, opened 2024-12-12
Redirect provenance/ to history/ and device/ to record/

bencoombs:
- Rename all links from provenance to history and device to record
- Add redirects for backwards compatability


RobertLRead (approved):
I tested this locally. I actually am not sure how the redirect works to support both urls, but it definitely works. I will merge it.

PR #376, opened 2025-01-14
Created Custom 404 Page

agilpatr:
Created the custom 404 page that now shows up instead of the default one.


RobertLRead (approved):
I tested this locally. I will merge it. Thanks, @agilpatr !

PR #377, opened 2025-01-14
Add send email function that uses Azure

bencoombs:
- Add function to send email using Azure SDK.
- Coco did dev and testing. Ben updated to latest main and added test.


RobertLRead (approved):
This looks great --- but I don't know how to test it!  @bencoombs can you point me to documentation on how to run the unit tests?  I think the fact that you created a unit test is fantastic...forgive me being unfamiliar with this---I see you are using "vitest".  Anyway, feel free to push this ---but email me instructions. Ideal I would test this locally, but I don't know how.

bencoombs:
The test can be run with 'npm run vitest' - currently this is separate to npm run test which runs Harry's integration test. At some point they will merge together.

This code is a bit isolated, but it will be called by the notifications logic.

PR #385, opened 2025-01-30
Notifications

jurrasicbark:
Implements a pub/sub style notifications feature.

- Added a REST API for subscribing and unsubscribing to notifications.
- Added a hook for publishing a notification.
- Created separate blob storage container which stores a list of a subscribers against ids. Uses the same encryption logic and metadata as record storage as the baseline for a common standard.
- Added notification types and business logic that can be extended to different notification integrations.
- Added unit tests.


RobertLRead (changes requested):
This is an impressive work. Your use of tests is good. I have a few questions or quibbles:

1. Should we have that default key at line 32 in notifications.ts? Is that a real key?
2. In line 153 of notifications.ts, does it it really unsubscribe if no tags are provided?  How do we subscribe to get a notification irrespective of tags (in essence, we want to subscribe to tags in disjunction (notify if t0  OR t1 OR t2 OR t3... etc. ) But then we need a special symbol to mean ALL_POSSIBLE_TAGS, so that one can subscribe to all tags. (This is rather the inverse of using the empty set with intersection.)
3. On line 165: Can you subscribe to different tags at different times? It seems like there might be a subtle bug here, where, if you are already subscribe to X, you can not subscribe to X OR Y in the future.
4. It looks like based on the function "getSubscribers" you are creating a new blob to store the subscribers for this key. That seems like a reasonable solution in the short term that we will want to change later. I'm surprised however that it looks like (I could be wrong) that you are storing a list of subscribers, meaning that every subscriber subscribes to the same set of tags?  That does not seem right to me---can you explain?

I am tempted to approve this to move the code base along, but I suppose you should comment before I approve it.  I don't know if subscribing to "all tags" and having specific tag lists for each subscriber is a problem for what you have written. If it is we can have a meeting to discuss it and possible at the same time discuss ways to prepare for an independent subscription system. For example, your code has given me the idea of publishing a JSON format for the subscriptions, which we could document, allowing third parties to participate. We could then implement our own third party notification system (but tell me that they should only use it if they trust us.)

But, if this works, it will be so valuable to DMDM that I think we should implement it as is immediately, and refactor later.

paulErdos:
We've taken a lot of inspiration from this as we develop the notification system. Closing for now, just to clear the draft from the list. We can re-open as needed.

PR #391, opened 2025-02-13
Update deployment

bencoombs:
- Integrate Azure communications services with staging
- Update build to multiple function files using modules
- Log the Nuxt configuration to check the deployments have the correct URLs set


RobertLRead (approved):
These look great!

PR #392, opened 2025-02-17
Qrcode blob

noramoor2007:
Added the view of the QR code into the record page:
Screenshot 2025-02-17 at 1 27 45 PM
Here is a preview of the hover effect.
download


RobertLRead (approved):
I approve this. I have not tested it. If you have tested it thoroughly, then go ahead and merge it.

PR #402, opened 2025-03-10
Limit description and title. Update packages.

bencoombs:


RobertLRead (approved):
I approve this, it looks simple

agilpatr (changes requested):
Looks good, but we should also implement the limit in CreateRecord.vue (this is what shows up in the asset history records page).

agilpatr (approved):

PR #404, opened 2025-03-10
Fix staging

### LINKED ISSUE #393: [BUG] Download Children Keys at CSV button does not download

bencoombs:
Fix bugs in staging - check for null, fix modal not rendering correctly, removed unused code.

Also closes #393

### DIFF
--- a/packages/frontend/components/Provenance/Feed.vue
+++ b/packages/frontend/components/Provenance/Feed.vue
@@ -50,10 +50,7 @@ along with this program.  If not, see <https://www.gnu.org/licenses/>. -->
 
         </div>
     </div>
-    <!-- The Modal to Enlarge Image -->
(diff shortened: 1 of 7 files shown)


agilpatr (approved):
@bencoombs Looks good! Two things, I can't click on the QR image to download it, but it seems like you should be able to due to the hover effect. I would recommend commenting out the hover effect on the image until the download feature is implemented (or removing it if we don't want to download the QR by clicking on the image, since the button does work).

Also, the QR code blocks important parts of the screen on mobile (including the record name and key). This feels like something that is important to address before we merge to the main site, and we should definitely create an issue for it, but for just merging into main on GitHub it should be fine.

RobertLRead (approved):
I approve this --- but honestly I don't understand it. I don't know what it is trying to do. I trust it is well motivated.

Issue #361, opened 2024-12-10
[BUG]  When you have a bad key, the message is not informative and there is an error on the page


RobertLRead:
**Describe the bug**

**To Reproduce**

Go here, which is a valid URL with an "x" appended:
[https://gosqas.org/provenance/8FFYE6suyVbCtVAWe2sP8Yx](https://gosqas.org/provenance/8FFYE6suyVbCtVAWe2sP8Yx)

It says it loads forever. 

Then go here:

[https://gosqas.org/provenance/8FFYE6suyVbCtVAWe2sP8Y](https://gosqas.org/provenance/8FFYE6suyVbCtVAWe2sP8Y)

And see the difrerence.

Note it thows an error on the page (bad promise) that you can see in the debugging tools:

![Screenshot 2024-12-09 at 7 54 31 PM](https://github.com/user-attachments/assets/81cfd0d7-e643-4e29-bf60-f122482d92f4)

**Expected behavior**
We should give an informative message saying "This is not a valid history key", and suggesting they go to a our home page, with a link provided.

**Screenshots**
https://gosqas.org/provenance/8FFYE6suyVbCtVAWe2sP8Y

**Desktop (please complete the following information):**
Chrome browser, from a Mac omputer.

RobertLRead:
@ninalahoti can you design a simple page for this and then assign it to Ben to assign to a developer?

bencoombs:
The device page catches that error so we could do the same on the provenance page?
![image](https://github.com/user-attachments/assets/ddd5dabc-aa9f-4c33-96f5-c90e0fb305cb)

ninalahoti:
https://www.figma.com/design/nFA3vMmmL6DdoOIpBmLgIC/Brand-Guidelines?node-id=2060-504&t=zNzSw9ufnJZ9s0LI-1 Thoughts on this page design and text? @RobertLRead @vjaqua

vjaqua:
One text edit:
In place of the word "email" just say the email gosqasystem@gmail.com

ninalahoti:
Made the edit! @bencoombs I believe this is ready to be assigned now

vjaqua:
This would resolve #386 and #357

bencoombs:
The tasks here are:
1. Create a component that renders the message and buttons that Nina has created in Figma.
2. Replace the text "Loading... please wait." with the new component. You can find this in packages/frontend/pages/history/[deviceKey.vue] on line 141.

PR #417, opened 2025-04-02
Add informative description for error page when record key is not found

suramdivyareddy:
PR for issue #361 

Changes in [deviceKey].vue
-Added an informative description and styling based on figma design.

After changes
image


RobertLRead (approved):
This looks real good --- it was simple, of course, but it looks good, thank you, @suramdivyareddy !

PR #420, opened 2025-04-14
Changing "Notify All Children" Language

agilpatr:
**Implemented “Annotate All Children” and “Recall All Children”:**
- Annotate All Children: Add a new record (just tags) to all children (same functionality as the old “recall” tag used to have, can modify to add descriptions as well if we want)
- Recall All Children: Add a new record (description and tags) to all children AND pin it to the top of the feed
- Choose both Annotate and Recall: Same functionality as “Recall All Children”

**Two things I wanted to clarify:**
- I did not add the red banner from the FIGMA, as the wording was confusing to me. A record in the feed was recalled, not the entire record feed. Also, it feels like it would clutter the feed to have a red pop up for every child (imagine having 200 children records all with a pop up every time you open them). I can implement this feature if we still want it, but I wanted to clarify the wording and logistics first.
- I’ve left “Notify All Children” on the Create Group page because it seemingly has no effect, the creation record is sent to all children regardless of if “Notify All Children” is checked or not. Unsure as to what effect we want this to have, or if it’s redundant and we want to remove it.


agilpatr:
Note: We are currently waiting on Dora's feedback, Victoria will reach out to her later this week to follow up.

RobertLRead (approved):

RobertLRead:
@agilpatr This has a lot of conflicts now (possibly because of me merging other PRs.)  Can you please resolve these conflicts and re-test?

I don't know if you know how to resolve conflicts---if you don't email me and we can do it together.

agilpatr:
@RobertLRead The conflicts should now be resolved, I tested them and everything runs smoothly on my end.

PR #424, opened 2025-04-22
Fixes #423.

### LINKED ISSUE #423: [BUG] Subtle difference between Windows and Mac created problem on Windows reading environment.
On a mac, if a key in the enviroment is not found, in typescript it is "undefined". On Windows, it seemed to be an Empty String (""). This was creating a difference between Rob and @JaraR 's cod that was making it hard for him to reach Azurite correctly.

We have fixed this by testing not just for "undefined" but for "undefined or empty".

RobertLRead:
@JaraR and I worked on this.

### DIFF
--- a/packages/backend/src/functions/httpTrigger.ts
+++ b/packages/backend/src/functions/httpTrigger.ts
@@ -220,8 +220,12 @@ async function convertLegacyProvenance(containerClient: ContainerClient, key: st
 }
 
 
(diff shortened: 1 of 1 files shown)


RobertLRead (commented):
I wrote this with @JaraR

RobertLRead (commented):
asdfasfd

Issue #299, opened 2024-10-07
[BUG] Clicking multiple times on the Create Device button creates multiple devices


bencoombs:
**Describe the bug**
Clicking multiple times on the Create Device/Container button creates multiple devices/containers. 

**To Reproduce**
1. Go to the home page
2. Open the console
3. Click on Create Device multiple times
4. See that there are multiple console messages reporting that the device was created

**Expected behavior**
Only one should be created.

PR #425, opened 2025-04-28
disabled the Create Record button when form was first submitted

deneille:
Stopped the submission of multiple devices/containers and from sending multiple console messages to report the creation of the device. 

To this, I disabled the submission button after the device creation form is validated and submit is clicked initially.

The issue this is related to is #299


RobertLRead (approved):
This looks good to me. I downloaded it locally and tested it locally. Sadly, I think when running locally it is so fast, that I could not really exhibit the bug that it it meant to fix --- which means I could not do a good "red/green" test. However, the code is clean and it works, so I am going to merge it.

Issue #419, opened 2025-04-11
[BUG] Formatting error w/new main branch


noramoor2007:
There is some weird formatting updates with the main branch, that has shifted some things visually on the asset history records page:

<img width="1512" alt="Image" src="https://github.com/user-attachments/assets/ed721464-bc76-49c4-a4ad-a29101655378" />

I remember adding the QR code but now it has been shifted down, overlapping other elements. Additionally, there was a share button on the record page that is no longer there and needs to be there for issue #319 

<img width="1512" alt="Image" src="https://github.com/user-attachments/assets/c9b6ff68-198d-42d9-961d-46c96ca55c1b" />

noramoor2007:
This is the before:

<img width="1512" alt="Image" src="https://github.com/user-attachments/assets/e3889a2b-c934-49ef-80e2-d81b8ca85a03" />

agilpatr:
@ninalahoti Here's some screenshots of the updated QR, I'm still working on implementing the Share Record Link button's functionality. This fixes Issue #413 (where the QR overlaps with other elements) so I'm not sure if we need that issue.

<img width="818" alt="Image" src="https://github.com/user-attachments/assets/3aa0f7ff-a78e-4f59-80cd-3d6c90575bc6" />

<img width="168" alt="Image" src="https://github.com/user-attachments/assets/36a9f9df-7752-4885-9175-ce2a12419d59" />

SereneBe:
@agilpatr @vjaqua I think the issue does overlap with bug #413 and this design does seem to match more with the figma pictured below. Should issue #413 be closed and I start working on implementing dark mode issue #350?

![Image](https://github.com/user-attachments/assets/7dd269cf-4ca2-44a8-9816-46ae27fd4fb5)

![Image](https://github.com/user-attachments/assets/d2c19d83-10bc-447f-8481-d583317cd892)

agilpatr:
@SereneBe I agree, I think you should focus on issue #350 for now. We can probably close #413, but we'll let @vjaqua make the final call there.

PR #426, opened 2025-04-29
The handIconBackground image path is now stored centrally in the nuxt…

noramoor2007:


agilpatr (changes requested):
Looks good! 2 things: first, we should probably remove the copy button (as that button has been redesigned and is being implemented in another issue [#419/#319]). Second, should we be making use of the handIconBackground variable in the index.vue file? I'm not sure I see the point of defining handIconBackground otherwise. Small note: I wrote the issue, but Ben was the one who pointed it out, so I'm not totally sure what he had in mind for this problem.

paulErdos:
@agilpatr Hey Alison, just going through all the PRs. No rush, but when this is done please move it to Ready For QA.

paulErdos:
@agilpatr Excellent. Really appreciate your clear communication and broad focus.

PR #431, opened 2025-05-06
Numbered entries

noramoor2007:
The entries are numbered, but there needs to be some clarification on how they should be numbered (i.e., backwards or like what is below?)
Screenshot 2025-05-05 at 8 22 35 PM


RobertLRead (approved):
I've reviewed this --- I  don't feel that I can do a good job of this, but I don't want to delay. So I'm doing it.

Issue #433, opened 2025-05-06
[FEATURE REQUEST] Need an api endpoint that returns the version of the backend


RobertLRead:
In order to be able to successfully tell if we have deplyed, we need a simple backend api endpoint that gives a version number which is compiled into the code.

It would be even better if this could print the Github commit of the version as well; there should be some way of doing that automatically so that people don't have to remember to update the backend version number.

The backend version number will have to be placed somewhere in the code, probably in httpTrigger.ts.

PR #435, opened 2025-05-08
added versioning endpoint to return the version, git commit and build…

deneille:
This returns the backend version via the API endpoint. The related issue is #433. 
Created an interface for the data (i.e., version, git commit, build time) that needs to be returned through the endpoint in versions.ts file. Created the endpoint by extending the httpsTrigger file and the updateVersion script to be run at buildTime and update the versionInfo fields. 

Updated the package.json file to run the update-version script, i.e. the updateVersion script.

To test versioning changing automatically can run: `npm run update-version`. This runs the `updateVersion` script but it changes the version in the `package.json` automatically. This is the following when the `http://localhost:7071/api/version` opened when running the backend:
Screenshot 2025-05-07 214659

N.B.: Curious about the value limit of each number in the version, i.e. `x.y.z` should each number in the version string have a limit?

### DIFF
--- a/packages/backend/package.json
+++ b/packages/backend/package.json
@@ -11,7 +11,8 @@
     "prestart": "npm run clean && npm run build",
     "start": "func start",
     "test": "ts-node ./test/test.ts",
-    "dev": "func start",
+    "dev": "npm run update-version && func start",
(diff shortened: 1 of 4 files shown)


RobertLRead (approved):
I tested this. Very impressive. Excellent work, A+, very well done, thank you.

PR #438, opened 2025-05-12
Fixed QR Formatting Bug

agilpatr:
Fixed issue where QR code would overlap other page components. Also worked with Nora to implement Share Button dropdown and functionality.


RobertLRead (approved):
I have tested this and it seems to work well. I am not sure what happened to the code I wrote, but I'm approving and merging this.

PR #441, opened 2025-05-15
Changed Email on Website to "info@gosqas.org"

agilpatr:
Changed the email to the new gosqas email and converted the email on the site to links (that open an empty email to info@gosqas.org).


RobertLRead (approved):
This looks good. I have not directly tested it, but it looks fairly simple. I am nervous about us using "hard" pixel values as you have done here:

```
    .link-font {
        font-size: 20px;
        line-height: 30px;
    }

```

I think it is better to use screen percentages, or sizes like "large". However, we have many examples of this "hard" value throughout our code.  Possibly I should open an issue for that.

Issue #350, opened 2024-11-23
Add darkmode to match Figma


bencoombs:

vjaqua:
https://www.figma.com/design/nFA3vMmmL6DdoOIpBmLgIC/GDT-by-GOSQAS?node-id=0-1&p=f&t=JY6rSRo9UpSaJvWs-0

PR #444, opened 2025-05-17
Added darkmode to deviceKey pages

SereneBe:
Issue #350


RobertLRead (approved):
I tested this and it looks pretty good, except for one image that I will add into an issue.

PR #448, opened 2025-05-27
Fixed Share Dropdown Bug

agilpatr:
Fixed the bug where a very long description would cause the share dropdown to no longer line up with the share button, as well as other formatting errors when the description was long.


RobertLRead (approved):
I've tested this and I am approving it, but I am very nervous of the "@media" statements and the very large number of "hard" setting of "px" values. I hope @ninalahoti knows a better way to do this. I personally have used Flexboxes to get effect in the past.

PR #459, opened 2025-06-03
Distinguished Key Types

agilpatr:
I created two commits for two different methods of distinguishing the keys. There are 4 types:
- Group Record Key
- Child Record Key
- Reporting Key
- Record Key

The method in the most recent commit changes the name "Record Key" to the corresponding name above based on the type on record.
Screen Shot 2025-06-02 at 3 42 09 PMScreen Shot 2025-06-02 at 3 42 17 PM

The other method (that I can revert to if it is preferred) adds an extra field in the description that says what type of key the record is (leaving the name "Record Key" alone).
Screen Shot 2025-06-02 at 4 11 23 PMScreen Shot 2025-06-02 at 4 11 31 PM


RobertLRead (changes requested):
Dear @agilpatr ,

This looks good, and I have tested it. However, I think you have missed the situation when a key is BOTH a group key and child key. That is, there is a 3-level system. There is a grandparent, there is a parent, and a child. The parent is the parent of the child but it is ALSO the child of the grandparent. I don't know how important this is.  Possibly we should just pick one or the other, but I think it would be clearer and say "Group key and child key".  Can you maybe address that?

agilpatr:
@RobertLRead @paulErdos I added the case you mentioned Rob, as seen in the screenshot below. I also found a bug during testing that I addressed, where the 'Group Key' section for solo records would not actually add the record to a group. That functionality should now all work.
Screen Shot 2025-06-18 at 3 27 43 PM

packages/frontend/utils/descendantList.ts
```diff
@@ -133,26 +133,39 @@ export async function addChildKeys(record: any, childKeys: string[], attachments
     }
 }
 
-export async function addToGroup(recordKey: string, records: any, attachments?: File[]) {
-    if (records[0].record.hasParent as boolean) {
+export async function addToGroup(childKey: string, groupKey: string, records: any, groupRecords: any, attachments?: File[]) {
```
paulErdos, line 136: Great indentation-saving function structure. Leading with error conditions is solid layout.

paulErdos:
One question for you, @agilpatr. When you say "That functionality should now all work", I presume you've tried it out on your machine and it works - is that the case?

Go ahead and merge main into your branch to resolve the conflict in [deviceKey].vue, check the functionality again to make sure the changes still operate as expected, and you're clear to merge this branch into main.

agilpatr:
@paulErdos Yes, I tried out the functionality and everything worked on my end. I just merged main into my branch and tested again, there should be no more merge conflicts and this branch should be ready to merge into main.

Issue #461, opened 2025-06-03
[FEATURE REQUEST] Update Child Keys CSV


SereneBe:
Child Keys CSV should display the reporting key, record key (parent), URL, and the name of object

PR #464, opened 2025-06-09
Children Keys CSV Update

SereneBe:
- Issue #461 
- added headers and listed the child name, parent record key, reporting key, & child key URL

ChildrenKeysCSV


RobertLRead (approved):
I tested and and it looks good, thank you.

PR #470, opened 2025-06-12
Updated DMDM Page

agilpatr:
Made the requested changes to the DMDM page, adding the new text and fixing the bullet point section.


RobertLRead (approved):
thanks for doing this quickly, @agilpatr

Issue #432, opened 2025-05-06
Update Provenance CSV to Highlight Reporting Key


deneille:
Reporting Key is currently within the Children arrays within the CSV and not highlighted as such. Plan: Create a Reporting Key column that stores the reporting key if exists and trim to the Children Names and Children Keys columns to drop the Reporting Key fields. 

This issue is related to Issue #211.

paulErdos:
Associated with this branch:
https://github.com/gosqasorg/asset-provenance-tracking/pull/473

PR #473, opened 2025-06-16
added column for reporting key

deneille:
Added a column for reporting key in the CSV for the provenance. Now the reporting key has a field of its own and is separate and distinct from the children keys. The following shows an example of it:
Screenshot 2025-06-16 173328.

This is related to Issue #432

### DIFF
--- a/packages/frontend/components/Provenance/ProvenanceCSV.vue
+++ b/packages/frontend/components/Provenance/ProvenanceCSV.vue
@@ -29,7 +29,8 @@ export default {
                 }
 
                 // Create CSV header
-                let csvContent = 'Timestamp,Description,Device Name,Tags,Children Names,Children Keys,Attachment File\n';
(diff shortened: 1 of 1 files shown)


packages/frontend/components/Provenance/ProvenanceCSV.vue
```diff
@@ -79,7 +91,7 @@ export default {
                         .map(name => `"${name.replace(/"/g, '""')}"`);
 
                     // Concatenate relevant data for csv file
-                    csvContent += `${timestamp},${description},${deviceName},${formattedTags},${childrenNames},${formattedChildrenKeys},${stringifyAttachmentName}\n`;
+                csvContent += `${timestamp},${description},${deviceName},${formattedTags},${formattedReportingKey},${childrenNames},${formattedChildrenKeys},${stringifyAttachmentName}\n`;
```
paulErdos, line 94: Needs indent

Issue #457, opened 2025-06-03
[FEATURE REQUEST] Record CSV Update


deneille:
This issue relates to the multi-level creation of keys.

In the Provenance CSV, we need to add parent key and url links to the CSV.

RobertLRead:
I am going to meet with @deneille  about this to explain, but want to put this down in writing, even though it is hard to understand. These edits should be made to (ProvenanceCSV.vue).

1. we do NOT need or want to print the children keys when exporting the provenance CSV. We have a separate CSV file for that, and I don't think the two should overlap.
2. However, if there is a reporting key (that is, a child flagged as a reporting key), then it should should be printed in the CSV.
3. The Key itself should be a column (which means it will be repeated in every row.) The Reporting key will likewise be a column, which will be repeated (either empty or present) in every row.) 

Note there are two other major issues (which possibly should be split off from this issue, I will talk to @deneille about that).

1. The provenance record should NOT allow child group keys to be added, because we don't want general users to be able to do that!  (This is a mistake which I should have caught). This may mean we need to add a new way of creating a grandparent structure after the fact; but it should not be the case that ANY user can add child or parent keys! So that needs to be removed, which may make it conditional in that component. (This is in PrevenanceCSV.vue)
2. When the child keys are exported (in "csvFile.vue'), we also want to have columns for:
A. The Parent Key
B. The Parent URL
C. The Device Name
D. The child Key
E. The child URL
F. A flag set to "T" on the row for the reporting key (this may not exist) or "F" if it is not the reporting key.

paulErdos:
@deneille Please use this branch for work under this ticket:
```
git fetch origin
git checkout 457-feature-request-record-csv-update
```

PR #478, opened 2025-06-26
updated the provenance and children csvs in accordance to Rob's desired outline

deneille:
In accordance to Rob's comment in Issue #457, the following changes were made to both the Provenance and Children CSVs.

Changes to the Provenance CSV:
- Device Key, Device URL and Reporting Key column were added.
- Children Keys and Children URL were removed.

Below is a screenshot of the updated provenance csv:
Screenshot 2025-06-25 221927
 
Changes to the Children CSV:
- The Parent Key, The Parent URL, The Parent Device Name, The Device Name, The child Key, The child URL were arranged respectively.
- The child URL was fixed.
- Added reporting key row back and instead added isReportingKey column that has a T flag when the row is the reporting key row and a F flag otherwise.

Below is a screenshot of the updated children key csv:
Screenshot 2025-06-25 224257

### DIFF
--- a/packages/frontend/components/Provenance/ProvenanceCSV.vue
+++ b/packages/frontend/components/Provenance/ProvenanceCSV.vue
@@ -29,7 +29,7 @@ export default {
                 }
 
                 // Create CSV header
-                let csvContent = 'Timestamp,Description,Device Name,Tags,Children Names,Children Keys,Attachment File\n';
+                let csvContent = 'Timestamp,Device Name,Device Key,Device Url,Description,Tags,Reporting Key,Attachment File\n';
 
                 for (const provenanceItem of provenanceData) {
                     // Format timestamp in UTC with both local and UTC time
@@ -48,19 +48,8 @@ export default {
                         .join(';');
                     const formattedTags = `[${tags}]`;
 
-                    //Find Children Names and format them
-                    const children = (provenanceItem.record?.children_name || [])
-                        .map(tag => `"${tag.replace(/"/g, "''")}"`)
-                        .join(';');
-
-                    const childrenNames = `[${children}]`
-
-                    //Find Children Keys and format them
-                    const childrenKeys = (provenanceItem.record?.children_key || [])
-                        .map(tag => `"${tag.replace(/"/g, "''")}"`)
-                        .join(';');
-
-                    const formattedChildrenKeys = `[${childrenKeys}]`
+                    //Get reporting key
+                    const reportingKey = provenanceItem.record?.reportingKey?.replace(/"/g, '""') || '';
 
                     // Get attachment filename
                     const baseUrl = useRuntimeConfig().public.baseUrl;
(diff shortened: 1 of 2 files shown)


packages/frontend/components/Provenance/ProvenanceCSV.vue
```diff
@@ -29,7 +29,7 @@ export default {
                 }
 
                 // Create CSV header
-                let csvContent = 'Timestamp,Description,Device Name,Tags,Children Names,Children Keys,Attachment File\n';
+                let csvContent = 'Timestamp,Device Name,Device Key,Device Url,Description,Tags,Reporting Key,Attachment File\n';
```
paulErdos, line 32: Rearrangement needed, per Rob: "Add the Device name as the third column."

packages/frontend/components/Provenance/ProvenanceCSV.vue
```diff
@@ -73,13 +62,16 @@ export default {
...
                     const stringifyAttachmentName = attachmentName
                         .map(name => `"${name.replace(/"/g, '""')}"`);
+                    
+                    //Stores device url and device key
+                    const deviceUrl = (window.location.origin + this.$route.fullPath).replace(/,+$/, '');
+                    const deviceKey = this.recordKey;
 
                     // Concatenate relevant data for csv file
-                    csvContent += `${timestamp},${description},${deviceName},${formattedTags},${childrenNames},${formattedChildrenKeys},${stringifyAttachmentName}\n`;
+                    csvContent += `"${timestamp}","${deviceName}","${deviceKey}","${deviceUrl}","${description}","${formattedTags}","${reportingKey}","${stringifyAttachmentName}"\n`;
```
paulErdos, line 74: Rearrangement needed, per Rob: "Add the Device name as the third column."

packages/frontend/components/csvFile.vue
```diff
@@ -27,26 +27,42 @@ export default {
...
                 // Skip the parent
                 const filteredChildrenKeys = childrenKeys.filter(key =>
-                    key !== this.recordKey && key !== reportingKey
+                    key !== this.recordKey
                 );
 
-                const csvRows = [['Child Name', 'Parent Record Key', 'Reporting Key', 'Child Key URL']];
+                let isReportingKey = ''; //Flag to check if row is the record key row or not
+
+                const csvRows = [['Parent Record Key', 'Parent URL', 'Parent Device Name', 'Reporting Key', 'Child Name', 'Child Key', 'Child Key URL', 'isReportingKey']];
```
paulErdos, line 40: Looking at isReportingKey and Rob's column list, it looks like we don't need / want the Reporting Key column.

packages/frontend/components/Provenance/ProvenanceCSV.vue
deneille: This should be fixed now. Thanks!

paulErdos (commented):
Looks great!

paulErdos:
Hey @RobertLRead: We're unclear on the definition of the reporting key: can we rely on the reporting key always being the child key of the first child?

deneille:
I've made the change for the Device Name column in the provenance csv. Here is an updated screenshot of the provenance csv with this change:
 
Screenshot 2025-06-26 214445

deneille:
> Hey @RobertLRead: We're unclear on the definition of the reporting key: can we rely on the reporting key always being the child key of the first child?

To add more to this comment, does the reporting key field still exist if not specified in the creation of the record? In the children csv method in the code, it specifies that if the reporting key does not exist then the first child becomes the reporting key. That logic does not exist on the frontend nor the backend. For example, in the backend this is what the getProvenance looks like with a reporting key:
Screenshot 2025-06-26 213836
and this is what it looks like without:
Screenshot 2025-06-26 213317.

The provenance with a reporting key has the reportingKey field whereas the one without does not.

RobertLRead (approved):
This is good. I have tested it. I made a minor tweak to remove the double quotes around the brackets in the tags, which seems to work better for me in both Numbers (Apple) and Excel (Microsoft), even thought that is not strictly a legal CSV value.

PR #481, opened 2025-06-30
Modified recall/annotate to work with multi-level keys

agilpatr:
Updated recall and annotate so they will now send records to all children in the tree, no matter how far down. Fixed test cases to make multi-level relations clearer, specifically made it so that a child in a group cannot be added to a different group (but its parent still can be). Also prevented the user from adding a group to itself.


packages/frontend/utils/descendantList.ts
```diff
@@ -169,54 +175,80 @@ export async function addToGroup(childKey: string, groupKey: string, records: an
...
                     postProvenance(key, {
                         blobType: 'deviceRecord',
                         description: "Annotated by admin",
                         children_key: '',
                         tags: tags,
                     }, attachments || [])
                 }
-                console.log("Finished updating children with 'annotate' tag.");
+
+                keysToCheck.shift();
```
paulErdos, line 208: Clever do-while.

agilpatr:
@paulErdos I just resolved the merge conflicts.

paulErdos (approved):
🚀

PR #496, opened 2025-07-06
Moved Background Image to Nuxt Config

### LINKED ISSUE #362: Move Background Image to the Nuxt Config
Move the hand icon background image on the front page (in the index.vue file) to the nuxt config, that way all the configuration is in one place.

agilpatr:
Updated code to store the background image in the nuxt.config.ts file. This performs the same as it does when we use url() in the css instead, and from what I found there doesn’t seem to really be a benefit to doing it this way.

I’ve created this as a PR, but I’m hesitant to merge it, since it seems to make the code more complicated than it started for no clear benefit. It also adds event listeners and new function calls which might ever so slightly worsen performance rather than enhancing it. @paulErdos Let me know what you think.


paulErdos:
@agilpatr I see what you mean. 

If there's no clear reason to prefer this method, then it's probably best we leave it out. 
That being said, if this is actually better for some reason (canonical / idiomatic, provides flexibility in ways we reasonably anticipate needing, etc), then perhaps it should be on the table. 

What's your opinion?

PR #516, opened 2025-07-14
Added Feedback When Create Button is Clicked

agilpatr:
Created a simple loading page (shown below) when creating records/groups, so that the user can clearly see that their records are being created. Also updated it so that the create record/group buttons will change color when clicked.

Screen Shot 2025-07-09 at 5 31 05 PM


agilpatr:
@paulErdos I think it looks a lot better with dark mode support for the loading screens (since they're seen every time a record is made). I just pushed changes to my branch that add dark mode to those pages, so this issue should be all set to merge now.

packages/frontend/pages/gdt.vue
```diff
@@ -75,12 +75,27 @@ along with this program.  If not, see <https://www.gnu.org/licenses/>. -->
 
         </div>
     </div>
+    <div v-else id="loading-screen">
+        <p class="text-center pb-5 pt-5">Creating record(s)...</p>
+    </div>
```
JaraR, line 80: I really like the use of v-if and v-else directives to conditionally render the loading screen based on whether or not the form has been submitted! It's a lot cleaner and less clunky to use when compared to React's conditional rendering.

JaraR (changes requested):
Approved the changes, but it may be better suited if the last part of the approval was requested as changes instead:
> A few issues appear when attempting to create a new record or group entry however. The loading screen only flashes for a brief period when a new entry is successful. If the entry is unsuccessful due to large files or poor/no connection, the loading screen does not appear while attempting to make that entry. The buttons also do not change color when clicked on new entries, which should be a quick fix. image

paulErdos (approved):
Works great! Thanks, @agilpatr!

paulErdos:
> Approved the changes, but it may be better suited if the last part of the approval was requested as changes instead:
> 
> > A few issues appear when attempting to create a new record or group entry however. The loading screen only flashes for a brief period when a new entry is successful. If the entry is unsuccessful due to large files or poor/no connection, the loading screen does not appear while attempting to make that entry. The buttons also do not change color when clicked on new entries, which should be a quick fix. image

Went ahead and tried this using a large 37mb photo of the milky way center: loading screen stayed up for me. We're going to merge this in for now, but if you can replicate the phenomenon let's get it noted down and filed under code quality improvements.

PR #518, opened 2025-07-15
Created Share Dropdown Component

agilpatr:
Created a component for the share dropdown (which appears on both the record and history pages). When looking over the code I didn't see any more duplicated code, at least not anything that was obvious.


paulErdos (approved):
Tested, works great. Thanks, @agilpatr!

PR #527, opened 2025-07-25
Python and Selenium scripts

HieuNgoGH:
Contains Python DOS files and polling script. Mostly work with Selenium.


paulErdos:
Hey what's up @HieuNgoGH. I'm seeing the folder, polling exe file, contains nearly 1000 files. We want to keep the memory footprint of the repo small, and keep it as simple as we can. 

Go ahead and decide whether those are needed: perhaps the executable strategy can be sidestepped by adding a requirements.txt so the user can install the needed dependencies. Another idea is to create a script to create the executable. If possible, go with the former.

paulErdos:
> Hey @paulErdos, thanks for pointing that out. I will remove the executable files and keep the scripts on there. I will also go with the requirements.

Epic. You got this! Let me know if you need any help.

paulErdos (approved):

PR #530, opened 2025-07-29
reducing the api calls when downloading children csv file

deneille:
This reduces the getProvenance api call to get the children keys and children names when downloading a Children CSV file.

Instead of calling the getProvenance api twice, it uses the original call to glean the children keys and children names.


paulErdos:
@Deneille Would you please add a screenshot of the issue no longer occurring either here or in the original issue?

paulErdos:
Screenshot_20250730_191212

paulErdos (approved):
🚀

PR #539, opened 2025-08-02
Fixed Error on Create Container Page

agilpatr:
A div was missing so the "customize grouped record titles" checkbox would give an error, is now fixed and works the same as before.


paulErdos:
Works. Text needs some styling cleanup which I'll take care of. Thanks, @agilpatr for figuring this one out!

paulErdos (approved):
Checked out and demoed: works.

PR #584, opened 2025-08-25
Real PR for Backend Unit Tests

agilpatr:
Added Backend Unit Tests. Currently most of them don't work, all the functions from httpTrigger.ts give a warning that skips over them (which is pretty much all of them). I commented out all the code that doesn't work, as well as leaving TODOs in httpTrigger and httpTrigger.endpoints.specs.ts about the warnings and where they're coming from. Since nothing is really being tested all tests do pass. Below are screenshots of the exact warning, as well as the tests passing.

Screen Shot 2025-08-28 at 5 45 57 PM

(I forgot to take a screenshot before commenting it out, but this warning would also appear in the github actions).
Screen Shot 2025-08-28 at 5 47 10 PM

As for blocking merge requests on test failure, it seems like this feature can be set up in the settings of the GitHub. Below is a Stack Overflow article, and while the question asks about rejection the top reply answers nicely in terms of blocking. I don't have access to this Github's settings, but it shouldn't be too difficult for someone with permissions to implement if we want.
https://stackoverflow.com/questions/58654530/how-to-auto-reject-a-pull-request-if-tests-are-failing-github-actions


paulErdos:
Thanks @agilpatr. I'm seeing those TODOs, and I appreciate you going above and beyond with regard to blocking. I'm also glad to see your use of commenting: you didn't let the integration tests get in your way, and you didn't let the apparent state of the tests get in your way. 

The warnings are fine: there is a lot of overlap in the names, but the warnings aren't related to the tests. This just happens when azure functions isn't being imported into an active runtime: basically the library is saying "I got imported, but I'm not mounting these edge functions to anything that would make them operational". And this is fine, because we mock the http interaction in httpTrigger.spec.ts on line 72. 

Noisy, to be sure: but their bark is worse than their bite. Very cool to see you succeeded in eliminating it by commenting the content of the tests as well as the import. 

This little slice of the screenshot you posted shows the tests passing. Un-commenting, I'm seeing they still pass. Which they should, they're mostly placeholders. 

Screenshot 2025-08-28 at 16 34 52

paulErdos (approved):
Great work, @agilpatr. This is going to get used a lot.

PR #600, opened 2025-09-03
Added User Manual Buttons

agilpatr:
Added buttons for the User Manual on the following pages:

Create Device/Container Page:
create_page

Learn More Section (the buttons shown are slightly different depending on the page, but all the pages have the user manual):
learn_more

Asset History Page:
history_page

For the Asset History Page, I wasn't totally clear on the style for the Download Child Keys and Download Provenance buttons, since they don't appear in the FIGMA. For now I just have the User Manual above them, but it would be good to get a set design for that at some point.


paulErdos (approved):
🚀

PR #615, opened 2025-09-05
434 feature request the front end needs a version number which can be reached from a url hieu

paulErdos:


HieuNgoGH:
Hey @paulErdos, the functions setVersion and getVersion have been tested and they're both working. 
434_set_get_version_working

Also side note, I was able to get VS Code's terminal to detach from the IDE and be its own window with its text enlarged separately.
VSCODE_terminalwindow_sizing

PR #619, opened 2025-09-08
Added keyFuncs service to backend

deneille:
- Copy keyFuncs.ts and keyFuncs.spec.ts from frontend to backend
- Adapt for Node.js crypto API instead of Web Crypto API
- Remove async/await since Node.js randomBytes is synchronous
- Maintain same functionality and validation logic
- Tests should run with existing Vitest setup


paulErdos:
@deneille Absolutely. Done. Lmk if that doesn't unblock ya.

PR #644, opened 2025-09-16
434 feature request the front end needs a version number which can be reached from a url hieu

HieuNgoGH:
setVersion and getVersion functions completed


paulErdos:
@HieuNgoGH, eres un chingon. You are killing it like Neo from the Matrix. 
Screenshot 2025-09-17 at 19 18 16

There's two things left, and no worries. See this area that talks about merge conflict and checks that pass or fail? Getting those clear is on you. 

# 1/2: How to handle merge conflicts

The merge conflict are a lil tricky at first. But once you get the hang of them they're very easy. Get help from me if you need. But the basics are this. 

When two people edit the same lines in the same file, git gets confused during the merge: which edit should be taken?

And it notates them like this

>>>>>>> (buncha greater than)
<code from one branch>
======= (buncha equals)
<code from other branch>
<<<<<<< (buncha less than)

And how I handle them is like this:
1. Read, decide which branch to keep, the one above or below the equals bar. Sometimes which branch is which is a lil confusing, no worries. 
2. Delete the one you don't want as well as the equals bar and the branch's corresponding bar of < or >. 
3. Delete the other bar of > or <. 
4. Good to go. 

Sometimes they can get a little squirrely. GitHub says it's "complex", but looking at the code it looks straightforward. 

# 2/2: How to handle checks. 

Sometimes checks just randomly fail for no reason. And they can be re-run.

Sometimes you can just re-kick them off by a button accessible in the test run (usually you can click on it but this one is being weird). However, always it is possible to re-kick the tests by pushing a new commit. 

I'll take a look into the tests and try to help you out.

paulErdos:
All right, check this one out. It's a tall section, and notice it has the format I was talking about with the <<< (code) === (code) >>> format. 

This one looks weird at first, but it's super simple: both of these functions got added to the same line. So all that's needed is just removing the merge conflict notation. I got this one for you. 

Screenshot 2025-09-17 at 20 45 27

paulErdos:
Screenshot 2025-09-17 at 20 49 07
Ah, there's two getVersions: Deneille's infrastructural placeholder, and your complete new one.

paulErdos:
Really stoked you wrote a test. It's failing, but not because the test is bad. It's self-destructing bc running on GH can be slow (it has to startup a container), and the default timeout is only 5sec. This got me, too. Here's the fix:

Take a look at the timeout in read.test.ts.

paulErdos:
Constructive feedback:

Take a look at this. The structure is a lil bit of a code smell that we can resolve nbd. I presume you looked through the docs for a "doesEntityExist" type function and found it missing. No worries. And honestly, props to you for your good functional way to handle it. 

Basically, the final catch block is for total complete failure. And we can shift it a bit, take a look. 

Add this structure:

```
export async function someThing(args) {
    try {                   // main try-catch

        <some code>

        try {               // Secondary try-catch!
            const response = table.add_new_thing() // might fail if exists
            return {some_json}
        } catch(error) { // bc entity exists
            const response = table.update_existing_thing() // technically could fail
            return {some json}
        } // End of secondary try-catch
       
        // -- Now, if an error occurs in the catch block of the secondary try/catch
        // it will get picked up by the final catch block. 

    } catch(error) {        // main try-catch
                            // main try-catch catastrophe section
        // Log the error internally
        console.error('detailed message')

        // For security reasons, be totally opaque to the user. Just inform them that a
        // generic error occurred
        return {
            status: 500,
            body: "Internal Server Error",
            headers: { "Content-Type": "text/plain" }
        }
    }
}

HieuNgoGH:
> Constructive feedback:
> 
> Take a look at this. The structure is a lil bit of a code smell that we can resolve nbd. I presume you looked through the docs for a "doesEntityExist" type function and found it missing. No worries. And honestly, props to you for your good functional way to handle it.
> 
> Basically, the final catch block is for total complete failure. And we can shift it a bit, take a look.
> 
> Add this structure:
> 
> ```
> export async function someThing(args) {
>     try {                   // main try-catch
> 
>         <some code>
> 
>         try {               // Secondary try-catch!
>             const response = table.add_new_thing() // might fail if exists
>             return {some_json}
>         } catch(error) { // bc entity exists
>             const response = table.update_existing_thing() // technically could fail
>             return {some json}
>         } // End of secondary try-catch
>        
>         // -- Now, if an error occurs in the catch block of the secondary try/catch
>         // it will get picked up by the final catch block. 
> 
>     } catch(error) {        // main try-catch
>                             // main try-catch catastrophe section
>         // Log the error internally
>         console.error('detailed message')
> 
>         // For security reasons, be totally opaque to the user. Just inform them that a
>         // generic error occurred
>         return {
>             status: 500,
>             body: "Internal Server Error",
>             headers: { "Content-Type": "text/plain" }
>         }
>     }
> }
> ```

Thanks for the constructive feedback @paulErdos. I didn't think about the nested try/catch! Should I redo the set/getVersions functions in this manner? If so I can get started on it asap. In regards to the read.test.ts, after pulling it to my system, should I add my tests in there?

HieuNgoGH:
Okay, I reread what you said in the constructive comment lol. Starting on it right now!

HieuNgoGH:
You're right, the GH tests do have to spin up the container and since it can't access what it would have locally, it's throwing errors.
image

HieuNgoGH:
@paulErdos Yippies! Removed the set/getVersion tests and also changed the setVersion structure. All checks have passed!

paulErdos:
All right, great progress @HieuNgoGH. Great to remove those tests: since it's trying to hit a live service, the tests needed for these will have to be integration tests. 

Overall, the integration test we'll want to do something like this:
originalVersion = getVersion()
const testVersion = // 
setVersion(bogusVersion)
const checkTestVersion = getVersion()
expect(checkTestVersion).toBe(testVersion)

Take a look at this integration test. We've got a <do api interaction> section and a <check that what we got back looks right> section. 
https://github.com/gosqasorg/asset-provenance-tracking/blob/e95c6acccebf1a5f50f9302c073a2087d3c31083/packages/backend/test/IntegrationTests/Live/read.test.ts

Have it hit Redstone, and what we'll do is, this is actually really cool:
1. Unit tests, 
2. Deploy to redstone,
3. Integration test against what just deployed during the workflow.

paulErdos:
Regex notes

https://github.com/gosqasorg/asset-provenance-tracking/blob/main/packages/backend/test/UnitTests/httpTrigger_no_mock.spec.ts#L53-L55

paulErdos:
Hey @HieuNgoGH, could you take another look at this?

PR #659, opened 2025-09-24
593 feature request combine key creation functions with getnewdevicekey endpoint

JaraR:
Key creation functions placed in the back-end. Testing also implemented and cleaned up. Verified to be working as expected.

image


paulErdos:
@JaraR Verified you have one add'l contribution logged under Insights > Contributors. Was 3, is now 4.

JaraR:
Nice one! @paulErdos 
`import { webcrypto } from 'crypto';`
`const { subtle } = webcrypto;`
I saw something similar last night but didn't think to apply it to webcrypto and subtle.

Back-end deployment was failing earlier this morning and it's working fine after a quick redeployment.

PR #683, opened 2025-10-07
Read: Feature Test : Read Group with Annotated Children 

HieuNgoGH:
Hey @paulErdos, let me know what you think!


HieuNgoGH:
Added some code to a previous that prevented passing the tests. Added a while loop to loop through all the record updates to scan for attachments because previous code only scanned one record update, not all.

image

PR #704, opened 2025-10-27
implemented a 2MB limit to the server side for attachments

deneille:
Implemented a size limit of 2MB to the server. This fix also handles most compressed files. Particularly files with extensions: `.zip`, `.gzip`, `tar.gz` and `gz`. 

To handle the compressed files, it streams the file and stops when it reaches the 2MB limit.

The following are POSTs to the `postProvenance` endpoint:

This is how a successful POST look like in the backend terminal:

FittedSize

This is an unsuccessful POST (or POST that goes over the attachment size limit):

OverSized

### DIFF
--- a/packages/backend/test/IntegrationTests/Live/create.test.ts
+++ b/packages/backend/test/IntegrationTests/Live/create.test.ts
@@ -639,7 +639,7 @@ describe("Group Creation Tests", () => {
 		
 		// Verify custom titles are present in all child keys
 		const verificationPromises = childKeys.map(key => 
-			fetch(`${baseUrl}/${key}`)
+			fetch(`${baseUrl}/provenance/${key}`)
 		);
 		const verificationResponses = await Promise.all(verificationPromises);
 		const verificationData = await Promise.all(


paulErdos (changes requested):
A couple things to change:
* I'm not sure it's necessary to unzip attachments. The size of what is sent is all that's needed. 
* I've thought a lot about it. I can't allow per-line edit history to get wiped for such an important file. Credit to all the contributors can't be squashed.

packages/backend/src/functions/httpTrigger.ts
```diff
@@ -163,6 +165,13 @@ async function uploadProvenance(containerClient: ContainerClient, deviceKey: Uin
     const attachmentIDs = new Array<string>();
     for (const attach of attachments) {
         if (typeof attach === 'string') continue;
+        // Check attachment size
+        let attachSize = attach.blob.size;
```
paulErdos: For this we don't want to trust the metadata for security reasons. It's still necessary to use the same technique of writing up to 2mb, and if it passes that, we delete and send back a 400 bad request. 

Same technique you were using before, just without the zipping (files come over the wire compressed already).

paulErdos:
@deneille, any idea what happened to the original commit that unzipped and counted bytes? I'm not seeing it here.

paulErdos (changes requested):
Take a look at the caller. We need to return 400 bad request rather than throwing an error. We definitely want to log the phenomenon as is happening here, but rather than crash, we want to be able to tell the frontend what happened so the user can be informed that an attachment is oversized.

PR #709, opened 2025-10-29
Clearer Record Creation Failure Message

agilpatr:
Created a function that will attempt to fetch three times before throwing an error (if it succeeds on any of those tries it will break out of the loop and return the response). This function has been implemented for both postProvenance (where Alex got the confusing TypeError) and getProvenance.

Below are screenshots of the error messages you could see. If the fetch() fails (aka. returns no error code, like what Alex was getting) then it will return a generic connectivity warning as seen below:
Screenshot 2025-10-29 at 6 19 16 PM

Otherwise the specific error code and default warning text will be displayed:
Screenshot 2025-10-29 at 6 20 21 PM

### DIFF
--- a/packages/frontend/services/azureFuncs.ts
+++ b/packages/frontend/services/azureFuncs.ts
@@ -24,26 +24,13 @@ export async function getProvenance(deviceKey: string) {
         
         const baseUrl = useRuntimeConfig().public.baseUrl;
         const fullUrl = baseUrl + "/provenance/" + deviceKey;
-        // TODO: can we get rid of underline (does work tho)
-        let response = await fetchUrl(fullUrl);
 
-        if (response == undefined) {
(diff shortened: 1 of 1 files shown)


packages/frontend/services/azureFuncs.ts
```diff
@@ -23,14 +23,14 @@ export async function getProvenance(deviceKey: string) {
         }
         
         const baseUrl = useRuntimeConfig().public.baseUrl;
-        const response = await fetch(`${baseUrl}/provenance/${deviceKey}`, {
-            method: "GET",
-        });
+        const fullUrl = baseUrl + "/provenance/" + deviceKey;
```
paulErdos: Very clean!

paulErdos (approved):

PR #724, opened 2025-10-31
Fixed Image Preview Bug

agilpatr:
Used @kdibba's code to fix the preview image not showing up, then modified the style so it looks the same way it used to. I tested with a couple different image and screen sizes and had no issues.

Screenshot 2025-10-31 at 3 44 17 PM


packages/frontend/components/Provenance/Feed.vue
```diff
@@ -270,10 +272,11 @@ export default {
     min-height: 300px;
 }
 
-/* TODO: remove? */
+/* TODO: remove unneeded styles (KEEP MAX HEIGHT! test on larger images when modifying) */
```
paulErdos: Good note taking.

paulErdos:
@agilpatr I really like the progression across commits. This is very good.

paulErdos (approved):

PR #731, opened 2025-11-05
Added Loading Screen to Asset History Page

agilpatr:
Added a few changes to make it clearer that a record is being created:
- Added a "creating record(s)" page when a record is being created and a "loading" page when the page is refreshed
- Added a success pop-up for when a record is successfully created
- Fixed error where loading screen would flash at the end of creating a record
- Made the page go back to the history page (with an error pop-up) when a record fails to create, rather than staying on the create page

Below are screenshots of what the page looks like when a record succeeds and fails to create (the creating page looks the same as the one from the previous issue):
Screenshot 2025-11-04 at 11 45 23 PM
Screenshot 2025-11-04 at 11 45 38 PM

A quick note: on PR #516 Jara mentioned an issue where the screen stays on the loading screen, even after record creation has already failed (From create device/group). At the time I didn't fully understand the implications of that, but after working on this I see how it could be confusing to see the loading screen when nothing is loading, especially if you miss the pop-up. I think it would be worth creating another issue to address this, as well as figuring out what we want that page to do instead (go back to the create device/group page? go to a new error page?). @paulErdos Let me know what you think and if I should create an issue for that.

### DIFF
--- a/packages/frontend/components/Provenance/CreateRecord.vue
+++ b/packages/frontend/components/Provenance/CreateRecord.vue
@@ -209,14 +209,14 @@ along with this program.  If not, see <https://www.gnu.org/licenses/>. -->
         },
         async submitRecord() {
             // Emit an event to notify the history/[deviceKey].vue page to display loading screen
-            EventBus.emit('isLoading');
+            EventBus.emit('isCreating');
 
             // Get a refreshed copy of the records
             let records;
             try {
                 records = await getProvenance(this.recordKey);
             } catch (e) {
-                EventBus.emit('isLoading');
+                EventBus.emit('isCreating');
             }
 
             if (!records || records.length === 0) {
--- a/packages/frontend/pages/history/[deviceKey].vue
+++ b/packages/frontend/pages/history/[deviceKey].vue
@@ -34,10 +34,10 @@ const recordHasParent = hasParent(provenance);
 <template>
   <!-- This link is for the icon in mobile dropdown menu -->
   <link rel="stylesheet" href="https://cdnjs.cloudflare.com/ajax/libs/font-awesome/4.7.0/css/font-awesome.min.css">
-  <div v-if="pageRefresh">
+  <div v-if="isLoading">
     <p class="text-center pb-5 pt-5">Loading record(s)...</p>
   </div>
-  <div v-else-if="isLoading">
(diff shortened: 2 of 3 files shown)


paulErdos:
@agilpatr 

re: A quick note: on PR https://github.com/gosqasorg/asset-provenance-tracking/pull/516 Jara mentioned an issue where the screen stays on the loading screen, even after record creation has already failed (From create device/group). At the time I didn't fully understand the implications of that, but after working on this I see how it could be confusing to see the loading screen when nothing is loading, especially if you miss the pop-up. I think it would be worth creating another issue to address this, as well as figuring out what we want that page to do instead (go back to the create device/group page? go to a new error page?). @paulErdos Let me know what you think and if I should create an issue for that.

Yes. This is an important thing to have found. I like the idea of returning to create device / group page. Adding retries that wait for 20x or response.ok may be worthwhile (perhaps as a wrapper so any fetching function can use it), although capping it would be best so as to account for, e.g., network outage. I like the idea of returning to the create record page.

agilpatr:
> @agilpatr
> 
> re: A quick note: on PR #516 Jara mentioned an issue where the screen stays on the loading screen, even after record creation has already failed (From create device/group). At the time I didn't fully understand the implications of that, but after working on this I see how it could be confusing to see the loading screen when nothing is loading, especially if you miss the pop-up. I think it would be worth creating another issue to address this, as well as figuring out what we want that page to do instead (go back to the create device/group page? go to a new error page?). @paulErdos Let me know what you think and if I should create an issue for that.
> 
> Yes. This is an important thing to have found. I like the idea of returning to create device / group page. Adding retries that wait for 20x or response.ok may be worthwhile (perhaps as a wrapper so any fetching function can use it), although capping it would be best so as to account for, e.g., network outage. I like the idea of returning to the create record page.

Great, I just created a new issue for that.

paulErdos:
Awesome work with the creating records(s) page and the loading screen.

How are you testing the record history update failing to create? I've tried a few ways. Running frontend and backend, creating a record, then killing the backend causes the page to immediately shift to the 404 page, without user interaction.

This isn't bad. But I am curious to see the error pop-up case. Reading the code I didn't happen to come across it, but perhaps I missed it. plz halp?

PR #734, opened 2025-11-06
697 devops backend workflow triggering incorrectly

deneille:
Added the negated paths to both `be_on_push.yml` and `fe_on_push.yml` files to ignore the respective directories when changes are pushed to the repository.


paulErdos:
Reviewed in Office Hours

PR #762, opened 2025-11-24
added integration grouped test for group record with zero children

deneille:
Since there is no true way to check the size of the `Number of Grouped Records`. This PR adds more test to the Group Record Creation, to be specific:
* A test to check when there are no children keys for a group record and,
* Verification that a bogus device key returns a 500 status code.

### DIFF
--- a/deploy_prod_fe.sh
+++ b/deploy_prod_fe.sh
@@ -16,5 +16,6 @@ export FRONTEND_URL=https://gosqas.org
 
 # Deploy frontend
 cd packages/frontend
+npm install
 npx nuxi build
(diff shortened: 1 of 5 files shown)


paulErdos:
@deneille Could you take a look at these merge conflicts? It looks like some totally okay code reuse is confusing git.

paulErdos (changes requested):
Check the test failures. There are a few different types, eg, 503 instead of 500, 60000 instead of 6000, and one that oddly is failing even with 60000 milliseconds. It's the large pdf upload. That one may genuinely need more time. 

Note the size on that pdf: if it's above two megs, replace it with a smaller one by the same name so these tests don't autofail as soon as attachment size limiting comes into action.

PR #769, opened 2025-12-01
update the api server so that swagger-ui docs point to the api server

deneille:
Updated the `translateAzureToSwagger.ts` file to permanently fix the api server problem.

The `test.sh` file runs and re-generates the yaml backend file and the json file in the frontend to add the respective api servers.

This fix allows a dtopdown where you can change to servers to run the docs locally, via staging environment and production. Here is how the dropdown looks currently:

Screenshot 2025-11-30 201404

Running the `translateAzureToSwagger.ts` also updated the `getNewDeviceKey` endpoint automatically with respect to the API endpoint of that operationId. Is this ok or should I revert this, @paulErdos?

Edit: So I reverted it by adding a manual override after re-reading the Interactivity ticket about avoiding hard work to be squashed during auto generation. 

New changes adds manual overridden path and checks for the path along with the `operationId` name of the path to avoid the path `device/Key` to be squashed in auto generation.

### DIFF
--- a/packages/frontend/nuxt.config.ts
+++ b/packages/frontend/nuxt.config.ts
@@ -50,7 +50,7 @@ frontendUrl: ${nuxt.options.runtimeConfig.public.frontendUrl}`);
     static: true
   },
   scalar: {
-    url: '/openAPI-docs.json',
+    url: '/openAPI-docs-with-servers.json',
     darkMode: true,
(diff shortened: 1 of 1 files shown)


paulErdos (approved):
Excellent work, Deneille!

PR #781, opened 2025-12-10
706 bug UI turbulence & DOM Manipulated Code

HieuNgoGH:
-Fixed UI Turbulence code that would submit form upon user pressing the 'Enter' key.
-Refactored DOM manipulated code


deneille (commented):
Thanks for this Hieu! It works for the most part except when checked, it still submits unless the user puts there cursor in the `Email` field. I am attaching two gifs the first with submits when I only click the checkbox (which we might want to avoid) and the other with is where your new commit is seen. Is it possible for you to handle it for the `checked`. I fou

paulErdos:
@deneille Excellent attention to detail. @HieuNgoGH Great responsiveness. 

Great work, you two!

PR #794, opened 2025-12-21
787 limit total blob size

### LINKED ISSUE #787: Limit Total Blob Size
Strictly speaking we are a key-value store. And we impose no limit on blobsize. Anyone could just store uncompressed DVD Video.ts files of lord of the rings extended edition in our pile of blobs if they wanted. We don't want copyright lawyers hitting us with a cease and desist right before a major release. And it's easy to think about how much worse this could get. 

In httpTrigger:
* Add a helper function and a function call to httptrigger that is used in creating new records and updating existing. 
* Whenever a record is created or an update comes in to an existing record, the record is checked before storage in initial creation or upon reading in case of update. If the new blob is larger than the limit or the edit would push the blob over the limit, return 304 not modified.

HieuNgoGH:
- Added middleware function that checks whether data coming through is under or over 2GB. This function is placed in postProvenance before data gets written to blob.


packages/backend/src/functions/httpTrigger.ts
```diff
@@ -268,6 +268,19 @@ export async function getDecryptedBlob(request: HttpRequest, context: Invocation
...
 }
 
+export function postProvenanceMiddleware(body): Boolean {
+    // This may seem simple but it is expected to grow
+
+    const sizeLimit: number = 2*10**9  // 2 gigabytes, this may change
+    var result = false;
+    
+    if (JSON.stringify(body).length < sizeLimit) {
+        return result = true
```
paulErdos: Just set result to true here for this coding pattern. 
Then there's only one return statement needed.

paulErdos (changes requested):
Great start. We also want to limit the total size of the blob itself, rather than just capping the size of additions. For example, the vulnerability left open here is that the blob size can be increased by 1Gb with each call. We need to look at the size of the blob being updated, and the size of the addition, and ensure the sum is smaller than the limit.

HieuNgoGH:
> Great start. We also want to limit the total size of the blob itself, rather than just capping the size of additions. For example, the vulnerability left open here is that the blob size can be increased by 1Gb with each call. We need to look at the size of the blob being updated, and the size of the addition, and ensure the sum is smaller than the limit.

Hey @paulErdos, I've added the functionality to check the total size of blob being updated (getProvenance) and ran into an issue with this line during unit testing.

https://github.com/gosqasorg/asset-provenance-tracking/blob/460b4f5e681190a276f5bc7109fa12ad9b733f3d/packages/backend/src/functions/httpTrigger.ts#L312

as you can see in the unittests results, that property is undefined during mocking but running the code locally with Azurite and Azure functions, everything is fine (records were created and appended successfully). I looked up the issue and a unit testing and mocking with the Azure SDK may circumvent this issue.

packages/backend/src/functions/httpTrigger.ts
```diff
@@ -287,7 +307,10 @@ export async function getProvenance(request: HttpRequest, context: InvocationCon
     }
 
     const records = new Array<ProvenanceRecord & { deviceID: string, timestamp: number }>();
+    let totalBlobSize = 0;
     for await (const blob of containerClient.listBlobsFlat({ prefix: `prov/${deviceID}` })) {
+        totalBlobSize += blob.properties.contentLength;
```
paulErdos: Looks like we've got a syntax error here. Cool to see GitHub itself catching errors!

paulErdos:
> > Great start. We also want to limit the total size of the blob itself, rather than just capping the size of additions. For example, the vulnerability left open here is that the blob size can be increased by 1Gb with each call. We need to look at the size of the blob being updated, and the size of the addition, and ensure the sum is smaller than the limit.
> 
> Hey @paulErdos, I've added the functionality to check the total size of blob being updated (getProvenance) and ran into an issue with this line during unit testing.
> 
> https://github.com/gosqasorg/asset-provenance-tracking/blob/460b4f5e681190a276f5bc7109fa12ad9b733f3d/packages/backend/src/functions/httpTrigger.ts#L312
> 
> image
> 
> as you can see in the unittests results, that property is undefined during mocking but running the code locally with Azurite and Azure functions, everything is fine (records were created and appended successfully). I looked up the issue and a unit testing and mocking with the Azure SDK may circumvent this issue.

Yeah, contentLength is a part of the BlobProperties interface, but it's not a required field.
https://learn.microsoft.com/en-us/javascript/api/%40azure/storage-blob/blobproperties?view=azure-node-latest

One thing Microsoft knocks out of the park better than most others is docs. The main docs list blobs is here: 
https://learn.microsoft.com/en-us/azure/storage/blobs/storage-blobs-list-javascript?tabs=javascript

The strategy to pivot to is to download the blob, and process the stream in chunks, and count the size of the chunks, and reject if over limit. 

Download docs are here:
https://learn.microsoft.com/en-us/azure/storage/blobs/storage-quickstart-blobs-nodejs?tabs=managed-identity%2Croles-azure-portal%2Csign-in-azure-cli&pivots=blob-storage-quickstart-scratch#download-blobs

I'll leave it a puzzle to you to find the rest. What you're looking for is how to download the blob in chunks. It's up to you to as to whether you'd like to look at these hints. Getting familiar with the docs is reeeaaaallly handy. Look around and see if you can find what you need. Like definitey spend some time looking around for what to do next to accomplish the download strat, but if you get stuck, spoilers below:
* Maybe useful but possibly not: https://learn.microsoft.com/en-us/javascript/api/%40azure/storage-blob/blobdownloadresponseparsed?view=azure-node-latest#@azure-storage-blob-blobdownloadresponseparsed-contentlength 
* Found it. Here's the way to do it, with docs all the way down:
    * Download as a stream we can find the download method docs taking us from here: https://learn.microsoft.com/en-us/azure/storage/blobs/storage-blob-download-javascript?tabs=javascript
    * to here: https://learn.microsoft.com/en-us/javascript/api/%40azure/storage-blob/blobdownloadresponseparsed?view=azure-node-latest#@azure-storage-blob-blobdownloadresponseparsed-contentlength
        * where we find a readable stream body
        * https://learn.microsoft.com/en-us/javascript/api/%40azure/storage-blob/blobdownloadresponseparsed?view=azure-node-latest#@azure-storage-blob-blobdownloadresponseparsed-readablestreambody
        * We know since the underlying datatype is not hyperlinked we're dealing with fundamental js; we're outside the ms library, so googling takes us to the official MDN docs: https://developer.mozilla.org/en-US/docs/Web/API/ReadableStream
        
Read up on readablestreams, then give the alternative strategy a shot.

HieuNgoGH:
Howdy @paulErdos ! I've fixed the merge conflicts as requested!

paulErdos:
Greetings, @HieuNgoGH! I was reviewing this PR, and I'm seeing that some of the postdeployment tests are failing. It seems like many are failing, however I suspect this may be a "from one source all thing depend" sort of phenomenon. Could you re-check the middleware (which is great to now have, it will make a good nesting place for other calls, e.g., email invocation) and see if you can identify the cause?

packages/backend/scripts/start-azurite.sh
```diff
@@ -4,4 +4,4 @@ if [ ! -d 'azurite_path.ignore' ]; then
     mkdir azurite_path.ignore
 fi
 
-azurite --location azurite_path.ignore &
+azurite --skipApiVersionCheck --location azurite_path.ignore &
```
paulErdos, line 7: Explicitly labeling this as OK: this is recommended by Azure staff and is at present an open problem under active development. See associated issue.

paulErdos (approved):
🚀

PR #796, opened 2026-01-03
Created Graphs for Recent Usage

### LINKED ISSUE #502: [FEATURE REQUEST] Graphic display of recent usage statistics
**Is your feature request related to a problem? Please describe.**
We want to be able to render an informative graph of the times that history records were created, for example, in the last week. This will be very similar to typical "usage" graphs or graphs "views" per hour.

**Describe the solution you'd like**
I expect "d3" or "plotly" to be used here. It should perhaps be on the statistics page.

**Describe alternatives you've considered**
A table driven approach could work, and would be a start---the hard part is probably collecting the data.

**Additional context**
This was split from Issue #113 .

agilpatr:
Created two graphs using plotly for the /statistics page:
- A graph that tracks the number of records created on each day of the week
- A graph that tracks the number of records created each hour over the course of a week

The graphs track all record entries (not just new records). I tested this on the gosqasbe backend, and the time taken to load only gets a few seconds longer, even with around 40,000 entries (screenshot below), so I don't think runtime should be an issue.

Screenshot 2026-01-02 at 5 29 29 PM


paulErdos (changes requested):
Work in progress, see other comment.

agilpatr:
Hi @paulErdos, I've updated this branch to use Azure Logs instead of Container Client. Some important notes:
- We're still using Container Client to get the total number of records/unique records. This is because Azure Logs only stores data from the past 3 months (which works for the other counts but not the total).
- I've updated our call to Container Client to filter out any blobs that are not records, as well as removed the code that looked at each blob, which greatly reduces the time it takes to load. My new code currently counts legacy records as records.
- Updated Azure Log queries to only count calls to postProvenance that succeeded (I mentioned in office hours the numbers not lining up between Azure Logs and Container Client, this change and filtering out attachments got the numbers to be the same).

To get an idea of time improvements, see the below comparison video. The new load time is a little less than a minute on average, and the old load time is a little less than three minutes. Both are pointed at production for this test (localhost lists less records since it filters out attachment blobs, but it still loops through the same amount), though in actual use localhost will look at the records on your local device. We're still waiting on getting the keys as environment variables which is why the backend tests currently fail.

https://drive.google.com/file/d/1QVKCKupKhBDWZhY03M4aAp9ii4UvLvZ7/view?usp=sharing

That was a lot of information, so please let me know if there's anything that needs clarifying!

paulErdos:
> Hi @paulErdos, I've updated this branch to use Azure Logs instead of Container Client. Some important notes:
> 
>     * We're still using Container Client to get the total number of records/unique records. This is because Azure Logs only stores data from the past 3 months (which works for the other counts but not the total).
> 
>     * I've updated our call to Container Client to filter out any blobs that are not records, as well as removed the code that looked at each blob, which greatly reduces the time it takes to load. My new code currently counts legacy records as records.
> 
>     * Updated Azure Log queries to only count calls to postProvenance that succeeded (I mentioned in office hours the numbers not lining up between Azure Logs and Container Client, this change and filtering out attachments got the numbers to be the same).
> 
> 
> To get an idea of time improvements, see the below comparison video. The new load time is a little less than a minute on average, and the old load time is a little less than three minutes. Both are pointed at production for this test (localhost lists less records since it filters out attachment blobs, but it still loops through the same amount), though in actual use localhost will look at the records on your local device. We're still waiting on getting the keys as environment variables which is why the backend tests currently fail.
> 
> https://drive.google.com/file/d/1QVKCKupKhBDWZhY03M4aAp9ii4UvLvZ7/view?usp=sharing
> 
> That was a lot of information, so please let me know if there's anything that needs clarifying!

Hey @agilpatr!
Regarding item 1. If containerClient takes >1sec to retrieve data, create a dataclass with a [time triggered](https://learn.microsoft.com/en-us/azure/azure-functions/functions-bindings-timer?tabs=python-v2%2Cisolated-process%2Cnodejs-v4&pivots=programming-language-javascript) setter method set to refresh, e.g., daily. This way, the state of member variables (the data) is in the backend at all times and is accessible by the dataclass getter without latency. 

Question here: what were the numbers you got in the profiling you did? Initial, current latencies?

Item 2: Nice! Non-record blobs are attachments. These are worth being displayed as another number tile on the statistics page. 

Item 3: Let's display (at least for ourselves) the failure count. Have visibility depend on whether or not the app is deployed to prod or dev (check env vars). If prod, don't display that datum.

paulErdos:
@agilpatr: seeing an uncommon error pop up in the build checks. Go ahead and loop that into your dev flow.

paulErdos:
Hey @agilpatr! Taking a look further into this, we have two options. The first is to add env vars to GitHub itself so the Actions runners can pick up the values. This would hit us in the db access bill every time a deployment runs.

The other option is to mock. An easy way to mock the get-token step would be to (in the actual code) put getStatistics > (presently) line 497, 498, 499 in a "getToken()" function, and then mock that function. This would be a lot of mocking.

For now I think it makes most sense to remove the unit test for getstatistics. It's not strictly speaking a unit test, as it refers out to external services. Whether we'll need a postdeployment integration test, a test that would consist of "make a call, see that it responds, see that what it responds with has the expected form" is needed is something we can expect to not need, and if we do need it, we'll know. There will be a lot of eyes our our dataviz.

agilpatr:
Hey @paulErdos, I'll get started on making those changes and ping you once they're done! As for your question, for production I got 15580 unique records and 22563 total records using my code. This is about 10000 less for both than previously reported on the statistics page. It used to take 3 minutes to load the statistics page and currently it takes 1. Not sure if that answers your question, so let me know if it doesn't.

agilpatr:
Hi @paulErdos, I made the changes you requested. Below is a screenshot of what the new dashboard looks like (with failure counts only appearing on development). I removed the unit test for getStatistics and made a class that will refresh the total counts once per day (as well as when the global class is first defined), rather than getting them every time getStatistics is called. I wasn't seeing a build error in the checks, could you specify what the error was and if we’re still seeing it?
Screenshot 2026-05-14 at 5 53 00 PM

paulErdos:
Hey @agilpatr! This is great progress. I did some more thinking / reading about the dataclass strategy. We may want to ultimately move this to storing the data in a single blob under a new blob directory: if a container spins up in a new region due to a request from a locale that isn't frequently accessed, the user will have to wait the full minute. We can leave this aside for now (unless you're feeling particularly ambitious) and alter it if we encounter anyone saying "hey, this is taking a long time". 

More importantly, there's one minor todo left. 

When running locally, I'm seeing this[1] in the terminal and this[2] in the browser. Looking into the matter, it appears this cannot be made to run locally. That being the case, we have two options:
1. Detect that we're running locally on the backend, and return dummy data, or
2. Detect that we're running locally in the frontend, not call the backend, and instead display a message, e.g., "This feature cannot run locally".

Essentially only those who have Azure access can get the creds needed for this to run locally. And that's fine. We should not hand out keys to everyone in case they want to work on this. But we do want to handle this error condition. 

Next steps:
* Choose 1 or 2. I personally think 2 is cleaner. Storing sample  data for local dev in its own file that is imported would work, but it would take extra work to update it if the page itself changes. 
* Add notes regarding this in the issue, and comments in the code. Basically we want to save devs who don't have azure access from having to duplicate time spent figuring out why it does not work locally. 

Great progress. This is moving forward well. 

Notes:
1. `[2026-05-21T20:40:12.641Z] Executing 'Functions.getStatistics' (Reason='This function was programmatically called via the host APIs.', Id=cf8dda06-41de-4985-a921-decfc3aaccf8)
[2026-05-21T20:40:12.693Z] Executed 'Functions.getStatistics' (Failed, Id=cf8dda06-41de-4985-a921-decfc3aaccf8, Duration=51ms)
[2026-05-21T20:40:12.693Z] System.Private.CoreLib: Exception while executing function: Functions.getStatistics. System.Private.CoreLib: Result: Failure
[2026-05-21T20:40:12.693Z] Type:
[2026-05-21T20:40:12.693Z] Exception: ClientSecretCredential: tenantId is a required parameter. To troubleshoot, visit https://aka.ms/azsdk/js/identity/serviceprincipalauthentication/troubleshoot.
[2026-05-21T20:40:12.693Z] Stack: CredentialUnavailableError: ClientSecretCredential: tenantId is a required parameter. To troubleshoot, visit https://aka.ms/azsdk/js/identity/serviceprincipalauthentication/troubleshoot.
[2026-05-21T20:40:12.693Z]     at new ClientSecretCredential (file:///Users/vincent/Repositories/asset-provenance-tracking/packages/backend/dist/src/functions/httpTrigger.js:67941:13)
[2026-05-21T20:40:12.693Z]     at getStatistics (file:///Users/vincent/Repositories/asset-provenance-tracking/packages/backend/dist/src/functions/httpTrigger.js:68481:22)
[2026-05-21T20:40:12.693Z]     at InvocationModel.<anonymous> (file:///Users/vincent/Repositories/asset-provenance-tracking/packages/backend/dist/src/functions/httpTrigger.js:8718:48)
[2026-05-21T20:40:12.693Z]     at Generator.next (<anonymous>)
[2026-05-21T20:40:12.693Z]     at file:///Users/vincent/Repositories/asset-provenance-tracking/packages/backend/dist/src/functions/httpTrigger.js:8567:77
[2026-05-21T20:40:12.693Z]     at new Promise (<anonymous>)
[2026-05-21T20:40:12.693Z]     at __awaiter (file:///Users/vincent/Repositories/asset-provenance-tracking/packages/backend/dist/src/functions/httpTrigger.js:8549:20)
[2026-05-21T20:40:12.693Z]     at InvocationModel.invokeFunction (file:///Users/vincent/Repositories/asset-provenance-tracking/packages/backend/dist/src/functions/httpTrigger.js:8716:22)
[2026-05-21T20:40:12.693Z]     at y.<anonymous> (/Users/vincent/Repositories/asset-provenance-tracking/packages/backend/node_modules/azure-functions-core-tools/bin/workers/node/dist/src/worker-bundle.js:2:39877)
[2026-05-21T20:40:12.693Z]     at Generator.next (<anonymous>).`
4. Screenshot 2026-05-21 at 14 26 19

packages/backend/src/functions/httpTrigger.ts
```diff
@@ -1131,6 +1237,16 @@ export async function createGroupHandler(request: HttpRequest, context: Invocati
...
+async function setStatisticsTotals() {
+    totals.setTotals()
+}
+
+// Once per day update the total record, record entry, and attachment counts
+app.timer('updateRecordCounts', {
+    schedule: `0 0 * * *`,
+    handler: setStatisticsTotals
+})
+
```
paulErdos, line 1384: Nice! Well done!

packages/backend/src/functions/httpTrigger.ts
```diff
@@ -37,6 +38,47 @@ const containerClient = new ContainerClient(`${baseUrl}/gosqas`, cred);
 
 const MAX_ATTACHMENTS_LIMIT = 1000;
 
+// Public class for getting statistics page totals
+class recordTotals {
```
paulErdos: Hey @agilpatr! I'm presently writing a ticket that will store and access additional data in this class. Could you please generify the name of the class as well as its instantiation, and then also de-generify the names of, e.g., setTotals to, e.g., setStatisticsTotals? Doing so here will decrease the likelihood of human error during renaming in this upcoming issue.

agilpatr:
Hi @paulErdos, sounds good! I'll get started on that.

agilpatr:
Hi @paulErdos, I've updated the code! The page will now skip the call to getStatistics if it's running locally and instead display the below message (the links go to the statistics pages of both dev and prod). I left some comments in the mounted() section to try and make it clear why it's skipped.
Screenshot 2026-05-23 at 4 21 45 PM

I also made the dataclass more generic by storing the totals in an object (that way you can just create a new function and set whatever names you want for the values). If you would prefer another solution please let me know.

agilpatr:
Hey @paulErdos, I was also able to update this branch to store the totals in a blob called 'statistics/totals' instead of the data class. The blob will create if it doesn't already exist, give default values of 0 if it for some reason can't update, and will now only update once per day (rather than having to also update values every time the file is setup). I removed the data class since this branch wasn't using it, but if you need me to add it back just let me know.

vjaqua:
Please prioritize this PR for the 6/13 Public Invention conference.

paulErdos (approved):

Issue #777, opened 2025-12-09
Create e-mail notification subscription on record creation form


JaraR:
Create a checkbox on both record and group creation form to subscribe to e-mail notifications. Make sure to include an empty text field for users to enter their email address. See #522 for figma mockups.

PR #802, opened 2026-01-12
Handing email address to backend func + unit test

### LINKED ISSUE #778: Clientside Application Logic: Hand the Email Address to the backend
This ticket is pure frontend logic, with no Presentation Layer / CSS stuff. We want to follow the example of the way feedback emails are handed from the frontend to the backend. 

Rather than parametrizing the existing postEmail function, create a new function. This function will need to be connected to both the createContainer and createDevice forms. 

For the backend endpoint to call, use 'notificationSubscription'. 

For this function, create a unit test in [this](https://github.com/gosqasorg/asset-provenance-tracking/tree/main/packages/frontend/test) directory. Follow examples from [this](https://github.com/gosqasorg/asset-provenance-tracking/blob/main/packages/backend/test/UnitTests/keyFuncs.spec.ts) backend unit test (note: not integration; this is more simple). 

For the test, it suffices to just make sure the API call fires. Feel free to mock the backend service if interested in formality, but this is a bonus.

These examples should be handy:
postEmail function:
https://github.com/gosqasorg/asset-provenance-tracking/blob/main/packages/frontend/services/azureFuncs.ts#L91-L104

postEmail being called:
https://github.com/gosqasorg/asset-provenance-tracking/blob/main/packages/frontend/components/Forms/CreateContainer.vue#L235-L237

in the submit form function:
https://github.com/gosqasorg/asset-provenance-tracking/blob/main/packages/frontend/components/Forms/CreateContainer.vue#L161

milenapetrov:
New function POSTing email to endpoint: /notificationsubscription + unit test file to verify API call fired.

Added a new file 'postNotificationEmail.spec.ts' -[in frontend/test directory]  to test new function: 'postNotificationEmail(email)' -[starting from line 106 in azureFuncs.ts]

- I'll use the new function in createContainer and createDevice once I complete issue #777 , which I realize should probably have been completed before this issue. 
- Aside: let me know whether different naming conventions are preferred/ if I've missed anything.


packages/frontend/services/azureFuncs.ts
```diff
@@ -103,6 +103,20 @@ export async function postEmail(email: string) {
     }
 }
 
+export async function postNotificationEmail(email: string) {
```
paulErdos: We also need the key. postProvenance shows what to do, a few lines up, line 70. And then just add that to the formData.

paulErdos:
> let me know whether different naming conventions are preferred/ if I've missed anything.

The naming conventions you're using look great so far. As for payload structure, we can write the backend to handle the format you decide.

milenapetrov:
@paulErdos 

*For when you get the chance*

Can you review my recent commits and verify I'm on the right track, whether this ticket and #777 should be consolidated, and what I need to ensure/alter to be merge ready. 

Thanks so much!!! ✨👾

paulErdos (approved):

paulErdos (changes requested):
Hey @milenapetrov! Pardon, I'd accidentally pressed "approve" when I'd meant to request changes.

paulErdos (approved):

PR #807, opened 2026-01-16
788 create backend endpoint to receive notification signup email address

paulErdos:


paulErdos:
Reviewed in Office Hours

PR #820, opened 2026-01-19
added robots.txt

CAI-yw:
Added robots.txt. This should be testable after it is deployed.

Still thinking about what other possible reasons there might be for the record leakage.

### DIFF
--- a/packages/frontend/public/robots.txt
+++ b/packages/frontend/public/robots.txt
@@ -1,0 +1,2 @@
+User-agent: *
+Disallow: /history/


CAI-yw:
Hi Vincent,

Thank you for replying.

When I ran the app locally, I was able to see http://localhost:3000/robots.txt.
Screenshot 2026-01-21 at 10 17 54 PM


I reran the frontend deployment workflow, but I saw that “deploy frontend” was skipped. I was wondering how I can trigger the deployment in a PR.
Screenshot 2026-01-21 at 10 18 11 PM


I checked the fe_on_push.yml file, does the frontend deployment condition mean that only pushes to the main branch will trigger a deployment?
Screenshot 2026-01-21 at 10 18 56 PM

Thank you!

PR #835, opened 2026-01-22
792 Detect Online/Offline

HieuNgoGH:
- Completed function to check whether user is online or offline
- Added two unit tests and added them to a new .spect.ts file under frontend/test. One to check whether the user is online and the other test to check for offline.

It was recommended to monkey patch the fetch function in place but that was a little difficult to implement so I went with passing the test function a bogus url.

### DIFF
--- a/packages/backend/src/functions/httpTrigger.ts
+++ b/packages/backend/src/functions/httpTrigger.ts
@@ -52,6 +52,8 @@ interface NamedBlob {
     blob: Blob,
 }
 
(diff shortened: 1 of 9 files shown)


HieuNgoGH:
Hey @paulErdos, Alison has reviewed and approved this pr but there seems to be a pending backend-run-unit-tests that is missing. I was thinking of doing a squash and merge via command line but wanted check with you on this first before I do.

PR #836, opened 2026-01-25
Annotate Feature Test

### LINKED ISSUE #790: [FEATURE REQUEST] API - Update: Feature Test - Annotate
Using the notifyChildren function in httpTrigger, create tests for the annotate feature (can use the current recall tests, also in update, as reference). Annotate should take a record from the parent and send it to all the children, replacing the description with a default message but sending the same tags as the original record (plus the annotate tag).

Note, we currently have some annotate tests in create.test.ts (though they don't use the notifyChildren function), it might be worth refactoring those tests to use the notifyChildren function, instead of writing totally new tests.

Here is what an annotated record looks like for a parent group:
Image

And this is what it looks like for a child record:
Image

agilpatr:
Created update feature tests for the annotate function. The test creates an annotation in a grandparent group and sends it to the child and grandchild records.

Also changed the baseUrl in our live integration tests to use gosqasbe instead of gdtprodbackend (as per Vincent's request). The only file I didn't do this with was read.test.ts because those records were hardcoded, and even though the records are only stored in gdtprodbackend they are still visible on the live site (see links below).
- https://gdtprodbackend.azurewebsites.net/api/provenance/FXFukdAGkkUzmC87G8vjZX
- https://gosqas.org/history/FXFukdAGkkUzmC87G8vjZX


paulErdos:
Hey @agilpatr, looks like your renames confused git and caused merge conflicts. Once those are cleared, this is ready to merge.

agilpatr:
Hey @paulErdos, those conflicts should all be resolved now.

paulErdos (approved):

PR #857, opened 2026-02-14
839 notify

CAI-yw:
When a record is updated:
- If there are users signed up for the record, those user(s) will receive an email notification.
- If no one has signed up for the record, the system will continue as normal.

Things that need to be set up before implementing/testing this:
- We need a from_email address (as shown in the [emailExperiment0](https://github.com/gosqasorg/asset-provenance-tracking/blob/7083a65cbe8b9322b3b346d072613a9f21a6c2c4/packages/backend/src/functions/httpTrigger.ts#L607-L634)
 function), which can be generated through Azure Communication Services.
- In the file local.settings.json, set up your "COMMUNICATION_SERVICES_CONNECTION_STRING".

Manual Test Results：
Attached are some screenshots of the manual test results:
1. Six email addresses were used for testing purposes.
emails logs

2.Some of the email notifications received.
email-2
4-4b5e-b5d7-c2b06fd92210" />

email-3

email-1


CAI-yw:
currently working on an unit test issue

PR #862, opened 2026-02-20
839-notify 

CAI-yw:
Notify users who subscribed to a record when it is updated. 

Please find the original PR here: https://github.com/gosqasorg/asset-provenance-tracking/pull/857


paulErdos (approved):
Reviewing in office hours. Approved, although one thing left to do before merging: find a way to have vitest skip the integration test that is attempting to send email

PR #871, opened 2026-03-02
799 implement offline banner on create pages

HieuNgoGH:
-Implemented offline banner component file under components (OfflineBanner.vue).

-Implemented offline banner into CreateDevice.vue, CreateContainer.vue, CreateRecord.vue, history/[deviceKey].vue files.

-Added global variable displayOfflineBanner (used camelCase in name instead of suggested snake case for code base consistency) in azureFuncs.ts and added condition change in offflineTestFetch function.

-Added v-if clause to OfflineBanner in the create pages to display or not display the banner based on its computed function displayBanner which grabs global variable displayOfflineBanner and returns true or false.

create_record_bottom
create_record_light
create_group_dark
create_group_light
history_dark
history_light

### DIFF
--- a/packages/frontend/services/azureFuncs.ts
+++ b/packages/frontend/services/azureFuncs.ts
@@ -13,149 +13,140 @@
 // You should have received a copy of the GNU Affero General Public License
 // along with this program.  If not, see <https://www.gnu.org/licenses/>.
 
-import { validateKey } from '~/utils/keyFuncs';
+import { validateKey } from "~/utils/keyFuncs";
 
 // Global variable used to control the display of offline banner on create pages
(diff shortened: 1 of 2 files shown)


HieuNgoGH:
> Banners look good! There's a duplicate try/catch block in offlineFetchTest on [line 179](https://github.com/gosqasorg/asset-provenance-tracking/pull/871/changes#diff-43cd60738ffc5e0a61f528dc5f4eeb4858b5886d49d30ade87d9f96b8dbb52b6L164), but once you remove that I think this PR is ready to merge.

@agilpatr Good catch! I missed that during a merge conflict. Fixed now, thanks!

PR #881, opened 2026-03-07
Empty Stash Function

agilpatr:
Created a function that will check every request in localStorage and fulfill them (in LIFO order). If a request fails to post it will return 404, if all records succeed or there are no records to post it will return 200. Keys that are successfully created are stored in localStorage under ‘gdt-cache-fulfilled’ so they can be displayed later.

I wrote 4 tests: stashing/removing 1 record, stashing/removing 2 records, removing when there are no requests, and removing when a record fails to create. A small note about the tests, the test file could not post FormData (something about tests not running from the browser and being in the frontend folder) so I used URLSearchParams for the tests instead. The emptyCache() function works for both test and regular use cases, and I tested the regular use case from localhost and found no bugs.

I also updated the cache requests tests to reduce reused code and prevent overlap between tests (by manually removing created records from the cache).

### DIFF
--- a/packages/frontend/test/unitTests/azureFuncs.spec.ts
+++ b/packages/frontend/test/unitTests/azureFuncs.spec.ts
@@ -120,35 +120,35 @@ describe('Tests to see if we can remove from the stash', () => {
     } as Response);
 
     const key = await makeEncodedDeviceKey();
-    let [formUrl, formData] = await createRequest(
-      key,
-      'Stored Record',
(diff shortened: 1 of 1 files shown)


agilpatr:
@paulErdos I updated emptyStash based on your comments, as well as fixing the failing test (the one that should pass even while dev is down, it was just timing out). Let me know if there's anything else you'd like me to change.

PR #921, opened 2026-03-22
Get PWA to Work

### LINKED ISSUE #902: Allow Users to Reach Creation Page While Offline
We want to figure out how to allow users to reach the creation pages even while offline. Currently unless the user was on a creation page before going offline they cannot access it, and after creating 1 record they lose access. This would likely be done by storing pages or even the site on the user's device so they always have access to it.

### LINKED ISSUE #931: See If PWA Can Work
While looking at PWAs for accessing the create pages while offline we realized two issues:
- The user had to go into menus to download the page (not prompted, hard to find)
- PWAs are heavily device/OS dependent (for example my Mac can only download PWAs from Safari)

We want to see if there's a way to overcome these issues and make this method work to easily reach the create pages while offline.

HieuNgoGH:
- Installed @vite-pwa/nuxt via 'npm install @vite-pwa/nuxt'.
- Set up service worker, pwa manifest, and added @vite-pwa/nuxt module in nuxt.config.ts file.
- Changed from localhost:3000 to '/' as the starting url.
- Can cache GOSQAS page to display create and other pages when there's no internet connection on desktop.
- Tried ngrok and Chrome dev tools to test mobile (Android and iOS) offline cache but to no avail. Documentation says service workers only work on https and local is http (with the exception of localhost on desktop browsers). With it working on desktop, I think the only thing holding back further testing is the lack of https.
- Tested on mobile and create pages are cached and accessible offline when https://dev.gosqas.org is saved to home screen as a progressive web app (pwa).

NOTE: deployed to dev.gosqas.org by @agilpatr to test offline capability and it works.

### DIFF
--- a/packages/frontend/app.vue
+++ b/packages/frontend/app.vue
@@ -1,16 +1,17 @@
 <template>
-    <div>
-      <NuxtLayout>
-        <NuxtPage/>
-        <NuxtSnackbar />
-      </NuxtLayout>
-    </div>
-  </template>
+  <VitePwaManifest />
+  <div>
+    <NuxtLayout>
+      <NuxtPage />
+      <NuxtSnackbar />
+    </NuxtLayout>
+  </div>
+</template>
 
 <style>
 @media (prefers-color-scheme: dark) {
   body {
-    background-color: #1E2019;
+    background-color: #1e2019;
   }
 }
 </style>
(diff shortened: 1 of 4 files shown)


HieuNgoGH:
Hey @agilpatr !, the commit conflicts have been fixed and was wondering I could get your review.  Since there are no blocking checks, either of us can just merge it into main. Thanks!

paulErdos (approved):

PR #934, opened 2026-03-28
extended createGroup 

### LINKED ISSUE #922: v2 Group Creation: Tags and ReportingKey

We'd like to extend this such that it incorporates the Tags and ReportingKey features that exist in the /gdt frontend page.

There are two parts to this:
B) Create integration tests

For the tests, go ahead and place them in the new v2 tests directory at backend/test/IntegrationTests/Live/v2. 
The tests themselves will be simpler than the v1 tests: hand a payload to a url, ensure the return type and format is correct, and check that the new group and its child records were created properly. 

Notes:
* The backend code for createGroup: https://github.com/gosqasorg/asset-provenance-tracking/pull/870/changes#diff-8631ba984e215261a69196d08659b43f126cf471708d82bdef41586da2bf5446R765-R926

deneille:
This extension incorporates:
-  tags + reporting key to the `createGroup` endpoint in `httpTrigger`, 
- and integration tests for create group with respect to tags + reporting key

NOTE: what's the difference between `title` and `deviceName`. I think they are basically one and the same.  If I could gather correctly from the code. Used `localhost:7071/api/` for testing, I don't believe the `createGroup` endpoint is as yet in development (or it was down when I was testing). This is also why it is not passing the deployment tests I believe.


paulErdos:
Hey @deneille, looks like some of the checks are failing.

packages/backend/src/functions/httpTrigger.ts
```diff
@@ -969,30 +969,43 @@ async function createChildren(context, number_of_children, tags?) {
 
         childrenKeys.push(thisChild)
         if(childrenKeys.length == number_of_children) { 
+            if(hasReportingKey){
```
paulErdos, line 1027: Hey @deneille! Just wanting to get some insight into your work. It's unclear why this is applied to just the last child, and why it's done as the break is occurring.

paulErdos: From Office Hours: 

The reporting key is a separate entire record that is part of a group.

paulErdos:
Testing locally (screen sharing in office hours) we're seeing the same tests fail locally as in the checks here. 

Specifically, these ones:
 FAIL  test/IntegrationTests/Live/v2/create.test.ts > Group Creation Tests > should create a group record with a reporting key
 FAIL  test/IntegrationTests/Live/v2/create.test.ts > Group Creation Tests > should create a group record with tags

There's another attachment-associated one that is failing locally: don't worry about that. Just these two. 

Upon the next commit, we'll see all tests fail: not to worry, dev is down as I'm actively working on the deployment bundler.

deneille:
Reverted `reportingKey` to be a part of `children_keys` in the backend. These checks pass. The check that does not pass is related to api `v1`

paulErdos:
> Reverted `reportingKey` to be a part of `children_keys` in the backend. These checks pass. The check that does not pass is related to api `v1`

Roger that. Thank you.

packages/backend/src/functions/httpTrigger.ts
paulErdos: Let's pull this from an environment variable.

packages/backend/src/functions/httpTrigger.ts
```diff
@@ -1015,14 +1015,22 @@ async function createChildren(context, number_of_children: number, custom_child_
         j++;
         childrenKeys.push(thisChild)
         if(childrenKeys.length == number_of_children) {
```
paulErdos, line 1025: Let's remove j entirely.

The wisdom here is that statefulness reduction is more valuable than O(n) over small data structures.

packages/backend/src/functions/httpTrigger.ts
paulErdos, line 1018: For later, not in this PR, Vincent to create ticket / or just do: move retries over to createChild

packages/backend/src/functions/httpTrigger.ts
paulErdos, line 1025: We can move the entire contents of this if statement outside the loop, right after the loop. This drops the break and the conditional.

packages/backend/src/functions/httpTrigger.ts
paulErdos: const baseUrl = process.env['backend_url'];

paulErdos, line 979: Due to multiple fetches, this entire function should house the three retries.

paulErdos: Really we should create a function that wraps fetch and tries 3x like fetchurl in the frontend (azurefuncs)

paulErdos:
Hey @deneille, can I have you resolve the merge conflicts on this branch? Thanks!

deneille:
@paulErdos <s>this is fully merged and the tests are passing except for one of Jara's. Not sure why because it passed on the local run of the tests.<s>. Ok found the issue uncommented the `Math.max` line and it worked. It's for test case 3 and it returns a negative number without the line.

paulErdos (approved):

PR #942, opened 2026-04-02
Added feature: Custom Record Titles

### LINKED ISSUE #919: v2 Group Creation: Add feature: Custom Record Titles

We'd like to extend this such that it incorporates the Custom Record Titles feature that exists in the /gdt frontend page. 

There are two parts to this:
B) Create an integration test

For the test, go ahead and create two new directories in the backend live integration tests directory: APIv1 and APIv2. Move the existing live tests into APIv1, and create a new test file in the APIv2 directory. 

The test itself will be simpler than the v1 tests: hand a payload to a url, ensure the return type and format is correct, and then check that the new group's child records correspond to the custom record titles requested.

JaraR:
From #919.

Extends the API v2 createRecord operation to include custom record titles for group children and includes integration tests for this.
Can be used like so:
```
curl -X 'POST' \
  'https://gosqasbe.azurewebsites.net/api/createGroup' \
  -H 'accept: application/json' \
  -H 'Content-Type: application/json' \
  -d '{
  "deviceName": "Group Title",
  "description": "Group Description",
  "number_of_children": 3,
  "children_name": ["a1", "b2", "c3"]
}'
```

Custom record titles currently operate in this manner:

- Custom record titles must be prepared as an array of strings. The "children_name" key will ONLY accept either an empty array or an array of strings as its value.
- Custom titles included in the children_name array will be used as child record deviceName during group record creation.
- children_name array can have less custom titles than the number of child records specified by number_of_children. This will create child records with custom titles only up to the number of titles specified, any other children will have blank titles.
- children_name can contain more titles than number_of children. This will create children with custom titles up to number_of_child and discard any extra titles.
- If there are number_of_children specified and no children_name key present, each child will be created with the parent's deviceName + an incrementing record number as their title.
~~-- A parent deviceName consisting of a completely empty string "" will create children with blank titles.~~
- Creating a group with a missing number_of_children key and a children_name array will not create children.

### DIFF
--- a/packages/backend/src/functions/httpTrigger.ts
+++ b/packages/backend/src/functions/httpTrigger.ts
@@ -959,23 +959,35 @@ async function createChild(context: InvocationContext, custom_title: string = ""
     }
 }
 
-async function createChildren(context, number_of_children, custom_child_titles?: string[], tags?) {
+async function createChildren(context, number_of_children, custom_child_titles, tags?) {
     const childrenKeys = []  // Named to correspond with metadatum name expected by frontend
     let thisChild;
     let custom_title;
-    // custom_child_titles.reverse()
+    let j = 1;
+    let parent_name;
+
+    if (typeof custom_child_titles === 'string' || custom_child_titles instanceof String) {
+        parent_name = custom_child_titles;
+    } else {
+        // custom_child_titles.reverse();
+        // parent_name = custom_child_titles;
+    }
+    
     for (let i = 0; i <= 3 * number_of_children; i++) {  // Re: 3 * num: three retries per; attempts are identical
-        if (custom_child_titles?.length > 0) {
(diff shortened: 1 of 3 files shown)


paulErdos:
@JaraR: Git gave me some trouble with this branch. I believe I've fixed everything, but I'd like to be sure. 

Without pulling or merging main into this branch, can I have you force-push from your 919 branch? There may be no differences, but I'd like to be sure. Thanks.

paulErdos:
Thanks for the force-push, Qui-Gon @JaraR. 

By the way, I really like how you handle many conditions in one fell swoop. I think this would be valuable for the rest of the team to see. If you'd like to show the changes in create.test.ts in the Monday meeting, let's do that. Also sending this to you on Discord.

paulErdos (approved):

PR #943, opened 2026-04-03
772 feature request make qr code download file name the child code example 4ad1q9vt2vjkrtbsiry8bppng

### LINKED ISSUE #772: [FEATURE REQUEST] Make QR Code DownLoad File Name the Child Code, example: "4aD1Q9vT2vJkrtbSiry8Bp.png"
**Is your feature request related to a problem? Please describe.**
A clear and concise description of what the problem is. Ex. I'm always frustrated when 
I can not tell apart down loads or worse when down loads may try to over write a previous down load.

The down loaded QR code came in with the file name "vqr.png"
Image

I would prefer it come in with the child number which is this case would look like "4aD1Q9vT2vJkrtbSiry8Bp.png"

This way I can save it to a folder and not have to think of a name or have to manage copy and paste to rename.

kdibba:
Made little changes in two files.


paulErdos:
Works. 
Screenshot 2026-05-06 at 22 00 27

paulErdos (approved):

PR #944, opened 2026-04-05
Group creation attachments

### LINKED ISSUE #924: v2 Group Creation: Attachments

We'd like to extend this such that it incorporates the attachment feature that exist in the /gdt frontend page.

There are two parts to this:
B) Create an integration tests

For the tests, go ahead and place them in the new v2 tests directory at backend/test/IntegrationTests/Live/v2. 
The tests themselves will be simpler than the v1 tests: hand a payload to a url, ensure the return type and format is correct, and check that the new group and its child records were created properly. 

Notes:
* The backend code for createGroup: https://github.com/gosqasorg/asset-provenance-tracking/pull/870/changes#diff-8631ba984e215261a69196d08659b43f126cf471708d82bdef41586da2bf5446R765-R926

CAI-yw:
Extended the createGroup endpoint so users can now upload attachments when creating a new group record. Backend code and one integration test was added for this feature.

A new integration test was added for this feature. The test covers creating a group record with attachments and verifies that the parent group record is created successfully, the child records are created successfully, and the uploaded attachments can be retrieved correctly.

A small frontend change was also made for attachment rendering.  Previously, when an attachment was a non-image file such as .txt or .pdf, the UI still attempted to render it as an image thumbnail, which caused a broken thumbnail to appear. Now, for non-image attachments, the UI will show the _file name_ instead of rendering a broken thumbnail.

Please let me know if you would prefer not to include the minor frontend rendering change, and also if you would like child record attachment uploading to be supported as part of this feature. 
Thank you.


paulErdos:
Hello @CAI-yw. Appears to be a minor oversight: some tests are failing. Once all checks pass, ping me and I'll give this a review. Thanks!

CAI-yw:
Hello @paulErdos ,

I tested this locally by setting the base URL to http://localhost:7071/api, and the test passes locally.

However, the Live/v2 tests in CI are hitting the shared deployed environment. It seems like the changes from my branch have not been deployed yet, so the deployed environment does not include the change that this test depends on. Because of that, the test may continue to fail in CI on this branch and possibly on other branches as well.

Could you take a look when available, or help deploy my change to the dev environment? Thank you!
Screenshot 2026-05-06 at 10 28 02 PM

paulErdos:
> Hello @paulErdos ,
> 
> I tested this locally by setting the base URL to http://localhost:7071/api, and the test passes locally.
> 
> However, the Live/v2 tests in CI are hitting the shared deployed environment. It seems like the changes from my branch have not been deployed yet, so the deployed environment does not include the change that this test depends on. Because of that, the test may continue to fail in CI on this branch and possibly on other branches as well.
> 
> Could you take a look when available, or help deploy my change to the dev environment? Thank you! Screenshot 2026-05-06 at 10 28 02 PM

Greetings @CAI-yw. You are correct: deployment happens only after merge to main. I've updated the workflow file to remove this condition. 

There is a separate issue in the way (unrelated to your work) that I'll take care of. In the mean time, thank you for the screenshot of the locally-passing tests.

paulErdos:
@CAI-yw: Update, work in progress. The deployment issue is solved in the local deployment script, the CORS issue is solved, I'm I'm working on applying the bundler to the CI/CD workflow. 

I'd continue working on this today. 

Note: I'm manually redeploying from the (operational) build script (as opposed to the work-in-progress CI/CD that is bundled). Re-running the only-failing final step will result in a false positive.

packages/backend/test/IntegrationTests/Live/v2/create.test.ts
```diff
@@ -0,0 +1,105 @@
...
+    expect(groupProv.length).toBeGreaterThan(0);
+
+    const groupEntry = groupProv[0];
+    expect(groupEntry).toBeDefined();
+    expect(groupEntry.record).toBeDefined();
+
+    // Verify group record attributes, and number of children record
+    expect(groupEntry.record.deviceName).toBe(payload.deviceName);
+    expect(groupEntry.record.description).toBe(payload.description);
+    expect(groupEntry.record.children_key.length).toBe(payload.number_of_children);
```
paulErdos, line 78: @CAI-yw Just a note, failing on this line.

paulErdos:
Hi @CAI-yw. We're now back in reality. The code is being deployed such that the postdeployment integration tests operate as expected, and the oversize deployment phenomenon is solved (and merged into this branch). We are back down to a normal amount of failing tests: one. 

Go ahead and prioritize this, and as needed, drop by an office hours and we can pair program on it to figure it out.

CAI-yw:
> Hi @CAI-yw. We're now back in reality. The code is being deployed such that the postdeployment integration tests operate as expected, and the oversize deployment phenomenon is solved (and merged into this branch). We are back down to a normal amount of failing tests: one.
> 
> Go ahead and prioritize this, and as needed, drop by an office hours and we can pair program on it to figure it out.

Hi @paulErdos , thank you for the help. The test seems to be able to pass now after Jara's commit/fix (related to baseurl in another function): https://github.com/gosqasorg/asset-provenance-tracking/pull/942/commits/f32e0e4ec45c3ae5128c6ebc9ff042890855b436

CAI-yw:
Hi @paulErdos, I’m working on the API spec for group record creation and wanted to double-check the desired API request format for `createGroup`:  
Should `createGroup` support only `multipart/form-data`, or should it support both `multipart/form-data` and the current `application/json` format (Right now, my implementation supports both)?

My main consideration is: `POST /provenance/{deviceKey}` for record creation seems to support `multipart/form-data` only. Should `createGroup` be consistent with that endpoint?


Below are examples of how `createGroup` could be called with different content types:

**`application/json`:**
```bash
curl -X POST \
  'https://gosqasbe.azurewebsites.net/api/createGroup' \
  -H 'Content-Type: application/json' \
  -d '{
    "deviceName": "Group Title",
    "description": "Group Description",
    "number_of_children": 3,
    "children_name": ["a1", "b2", "c3"]
  }'
  ```

**multipart/form-data without attachments:**
```  
curl -X POST \
  'https://gosqasbe.azurewebsites.net/api/createGroup' \
  --form-string 'groupRecord={
    "deviceName":"Group Title",
    "description":"Group Description",
    "number_of_children":3,
    "children_name":["a1","b2","c3"]
    }'
```
  
**multipart/form-data with attachments:**
```
  curl -X POST \
  'https://gosqasbe.azurewebsites.net/api/createGroup' \
  --form-string 'groupRecord={
        "deviceName":"Group Title",
        "description":"Group Description",
        "number_of_children":3,
        "children_name":["a1","b2","c3"]
        }' \
  -F 'attachment=@./test/attachments/a200.jpg;type=image/jpeg'
```
  
Supporting only `multipart/form-data` for createGroup would allow both attachment and non-attachment use cases.
Right now the code supports both multipart and JSON. If we decide to support only multipart/form-data, some minor adjustments will need to be made in createGroupHandler, the tests, and the API spec. Thank you!

Reference: https://github.com/gosqasorg/asset-provenance-tracking/pull/942

**Comment by paulErdos** [comment]
Hey @CAI-yw, following up from our conversation. Taking a look at the [spec](https://github.com/gosqasorg/asset-provenance-tracking/blob/main/packages/frontend/public/openAPI-docs-sample.yaml), it would indeed be cumbersome to support both types. 

From what we saw in the devtools network tab, that being that regardless of attachments, the request is sent with the multipart/form-data, go ahead and support that case only.

File: packages/backend/test/IntegrationTests/Live/v2/create.test.ts [hunk 2]
```diff
@@ -22,6 +23,88 @@ describe("Group of tests", () => {
...
+
+    // Verify group attachments
+    expect(groupEntry.attachments).toBeDefined();
+    expect(groupEntry.attachments.length).toBe(2);
+
+    const attachmentHashes = groupEntry.attachments;
+    const imageDownloadResponse = await fetch(`${baseUrl}/attachment/${groupKey}/${attachmentHashes[0]}`);
+    expect(imageDownloadResponse.ok).toBe(true);
+    const downloadedImageBuffer = Buffer.from(await imageDownloadResponse.arrayBuffer());
+    expect(downloadedImageBuffer.equals(imageBuffer)).toBe(true);
```
paulErdos, line 88: Nice.

paulErdos:
Well done, @CAI-yw. This is very good. Nice attention to detail on that last commit. 

Running the tests locally, I'm getting 400 bad request on line 53. I'm re-running the tests. I'm seeing that the endpoint does exist in the deployed dev backend.

Ah. The changes, though, are not. This should auto-deploy, and says it did. Looking into it.

paulErdos:
Backend deployment job was not auto-triggering on "re-run all jobs". Manually running that one did it. Added a global timeout to some of the v1 tests to clean up transient timeout failures.

paulErdos (approved):

PR #951, opened 2026-04-08
PWA User Guide

### LINKED ISSUE #938: Create User Guide for the Offline PWA
Now that we've confirmed the PWA approach to storing the web app works on iOS and Android, we'll need a 1-pager guide showing users what to do to set it up, e.g., on iOS add to homescreen, create record, and then everything is set up. 

We'll want to note that users should check that offline mode is set up before leaving a service area. 

We'll want to be thematically consistent with the user manual. Go ahead and reach out to Christina to get a link to copy the format.

HieuNgoGH:
Created a docs folder inside package/frontend and placed the Offline Mode progressive web app user guide in there.

NOTE: "Binary file not shown" when viewing Files changed. Select the three dots (...) option --> View file and the pdf file will display.


agilpatr (changes requested):
Looks great, love the pictures and highlights! Two small things, in the Android with Firefox section you mention clicking the three dots but highlight a different symbol, so if you could either update the highlight or mention the other symbol that would be great. And in both the Android with Firefox and the Android with Chrome section under the 2. it should say "To test offline capability, turn on airplane mode" not online.  Otherwise looks good! Ping me after updating those two things and I can merge this in.

Image

HieuNgoGH:
> Looks great, love the pictures and highlights! Two small things, in the Android with Firefox section you mention clicking the three dots but highlight a different symbol, so if you could either update the highlight or mention the other symbol that would be great. And in both the Android with Firefox and the Android with Chrome section under the 2. it should say "To test offline capability, turn on airplane mode" not online. Otherwise looks good! Ping me after updating those two things and I can merge this in.
> Image

Good catch and thanks @agilpatr !, I've made the changes as requested. I also added mode "To test offline mode capability, turn on airplane mode" to better describe the capability.

agilpatr (approved):
Looks great!

PR #952, opened 2026-04-08
Python ddos script with parameters

### LINKED ISSUE #884: Parametized DDOS Python Script
This is an updated version of our DDOS script written in Python incorporating Selenium ran in command line. It allows us to simulate DDoS attacks through single IP and multiple-IP (IP spoofing) methods. The first two parameters are required with an optional third parameter that turns the IP switching feature on. The first parameter takes the number of requests and the second parameter takes in the number of bots to run. Because Github doesn't allow posting secret keys, the free access key acquired through an AWS account will be omitted from the uploaded file.

Example to run file:
When in the terminal, go to the directory containing the python script, type in: python ddos_with_parameters.py 5 5 on
This will run python bot script with 5 requests total, 5 bots, and IP switch on
Third argument can be blank and the script will run in single IP mode

HieuNgoGH:
Added Python ddos attack script with parameters taken through CLI to the backend/scripts folder.


paulErdos:
Hi @HieuNgoGH! Seeing a confusing result: many tests failing. 

Likely unrelated: go ahead and put this in backend/scripts. Its inclusion in the test hierarchy may be what is confusing the tests, although I doubt it. Regardless, this is more a util we'll want to use on occasion rather than something we'll want integrated into our test workflow (although we later will want to use selenium to validate ui workflows).

paulErdos:
Re-ran the one failed check, and it failed again. Very odd! See what you can do.

HieuNgoGH:
> Re-ran the one failed check, and it failed again. Very odd! See what you can do.

Hey @paulErdos!, I moved the file to backend/scripts and fixed the issue that was causing the failed test.

paulErdos (approved):

PR #973, opened 2026-04-14
Moved createRecord Into Backend

### LINKED ISSUE #896: Move createRecord into the backend
To get familiar with the nature of this task, read the description in this issue's parent (#894). 

We'd like a backend endpoint for record creation. Much of the code is already written. So let's first take a bird's eye view of what to do:
* Set up the endpoint in httpTrigger, along with a handler function to be invoked when the endpoint is hit
* Create a standardized zod schema so that we don't process malformed input, and can inform the user if they need to check their payload
* Drawing from the non-group record creation code in [this](https://github.com/gosqasorg/asset-provenance-tracking/blob/main/packages/backend/test/IntegrationTests/Live/create.test.ts#L1074) file (starting around line 1074 at time of writing), create a function or collection of functions to be called by the handler that do the operation of creating a record. Note: don't hardcode urls, rather follow the example of createGroup's use of environment variables. 
* In the live integration tests directory, in a new file, create a simple integration test that creates a record, gets it, and checks that it has the parameters with which it was created. Note: follow examples in the existing live integration tests.

The goal for this ticket will to have all of the features work. Start with title and description, and then progressively add features from the integration test code. 

* (Note: if you need additional environment variables to exist in deployment, ping me and I'll set them up.

agilpatr:
Moved createRecord functionality to the backend. The endpoint can handle all basic record features (name, tags, attachments, etc) and has a zod schema to check the given format. The only two features that don't currently exist on the endpoint are receiving email notifications and providing feedback. I followed the example of createGroup so it's set up very similarly.

For attachments they had to be converted to a base64string before being sent (the other option was to use formData as we had previously, but I decided against it since that would make this function very similar to postProvenance).

writeAttachment test was failing so I updated it to use a file smaller than the file limit.


paulErdos:
Hey @agilpatr, missed a line.

By the way, keep in mind, "PR opened" is not the final step. If checks aren't passing, things can't come in. It's the dev's responsibility to make sure a contribution is ready to merge. 

The way I do it is this:
1. Open PR
2. Wait for checks to run
3. If checks don't pass, then I have more to do, and I might convert back to a draft
4. If they do pass, then I deallocate my attention and go do something else. 

There's a trope of engineers "hurling untested code over the wall to be someone else's problem", and I know that's not you.

agilpatr:
Hi @paulErdos, so sorry about that! I'll try to be more careful going forward. The tests are all passing now, let me know if anything else needs to be changed on this branch.

paulErdos (approved):

PR #978, opened 2026-04-19
Created Rate Limiter

### LINKED ISSUE #976: Create Rate Limiter
We’ve come up with a rate limiter. We look at usage quantity over a timedelta, and if overthreshold, we decide to politely decline the interaction (http 429 too many requests). 

However, running a query for each API interaction can run us into the millions of data inspected in a short period of time. 

The solution is to use Azure time trigger functions. These are auto-invoked every timedelta (configurable). 

The time-triggered function will stick a label on a signpost, ie, set a global boolean as to whether we are receiving too many requests. 

All functions that do API interaction will then refer to the state of this variable. If false, continue as usual, if true, we respond 429. This will be received by the frontend, and a message can be displayed to the user.

agilpatr:
Created a rate limiter for our most used API calls. This rate limiter runs every 5 minutes and will check the number of calls made to each function. If the number of calls is over the limit it will set the flag for that function to true, which will prevent anymore calls to that function until the next run of the rate limiter (which will set the flag back to false). I checked each page that has rate limiting applied to it and they all display the error "429 Too Many Requests" when the limit is reached.

For this to work on dev we need to add some keys to Azure as environment variables. I wrote up steps for how to do this here: https://docs.google.com/document/d/1M7RmAOPpCI9SO1jn8T4bc7l-zkznuBtPHHhx_i32qCM/edit?tab=t.5rfjlnk4y6qf. Please reach out to me through discord or email if you run into any problems.

(Creating this as a PR even with azure stuff unfinished so it can be merged without me later if needed. Code-wise this should be complete unless there are changes requested).


agilpatr:
Note: I updated my code to rate limit all API calls that aren't within another call. Also the one failing test is a known issue and has a fix in [this pr](https://github.com/gosqasorg/asset-provenance-tracking/pull/955).

PR #979, opened 2026-04-20
Scan QR Code While in App or Website to Extract Device Key Without Exiting

### LINKED ISSUE #939: Scanning QR While Offline Takes You Out of PWA
When you scan a GOSQAS qr code from a mobile device while offline it will by default take you to Safari (or whatever your default browser is) rather than the downloaded PWA, causing it to fail. We want to find a way to scan qr codes and have them try to open the history page in the PWA instead.

HieuNgoGH:
Note: Although the feature works, it needs a real home in one of the pages in offline mode. Until then, it'll reside here and the code can be transferred over once the page is up.

- Implemented code to allow user to open up their camera while on GOSQAS website or PWA app without leaving said site or app.
- The video contents will be drawn onto a canvas so data extraction can be made.
- The camera will scan for a GOSQAS qr code and its data will be captured and parsed by jsQR. RegEx is used to extract the device key from the website on qr code. The key can be used to add additional updates to it in offline mode.
- The key will be posted on the site (feature for now so user can copy and paste the device key).
- Camera will close after successful qr code scanned.

### DIFF
--- a/packages/frontend/package-lock.json
+++ b/packages/frontend/package-lock.json
@@ -14,6 +14,7 @@
         "bootstrap": "^5.3.3",
         "bootstrap4-toggle": "^3.6.1",
         "bootstrap5-toggle": "^5.1.1",
+        "jsqr": "^1.4.0",
         "mitt": "^3.0.1",
         "nuxt": "^3.11.2",
         "nuxt-snackbar": "^1.1.2",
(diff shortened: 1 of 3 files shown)


agilpatr (approved):
Hey Hieu, this all looks great! I've commented out the qr scan button for now and left a todo to move it into the offline history page (something I'll handle in my own branch, so no need to worry about it). Next offline meeting we'll all figure out where to put this button long term.

PR #984, opened 2026-04-24
983 diagnose and treat typescript compiler issues

### LINKED ISSUE #983: Diagnose and treat Typescript compiler issues
We have had issues in our code that violate the typescript language. 

The Typescript compiler itself has erroneously not flagged them. Github user bakkot sits on the ECMAScript Technical Committee commented in [this](https://github.com/microsoft/TypeScript/issues/62546) issue in Microsoft Typescript repository:

"Previous versions of TypeScript incorrectly failed to give an error on your example, but they always should have."

This has caused the appearance of issues like those illustrated below that cause the backend to fail to launch. Note especially `Type '"SharedArrayBuffer"' is not assignable to type '"ArrayBuffer"'.`:

Image

paulErdos:


JaraR (approved):
LGTM

Issue #850, opened 2026-02-11
Periodic Checker


agilpatr:
We need a function that will be called once the code detects the user is offline, periodically checking if the user has come back online, then fulfilling requests once they are. Pseudo code of this functions layout is shown below (connectivityChecker and emptyStash are sub-issues of this one, so don't worry about writing those functions).

```
Def periodicChecker() {
   // If a process is already running, return (prevents multiple processes)
   If (localStorage[‘gdt-awaiting-connectivity’] == true) { 
      return 
   }
   localStorage[‘gdt-awaiting-connectivity’] = true

   stash_empty = false
   while (!stash_empty) {
	await connectivityChecker()  // wait for user to come online
	response = emptyStash()  // empty as much of the stash as you can
	if (response == 200) { stash_empty = true }  // if stash is empty, leave loop
   }

   localStorage[‘gdt-awaiting-connectivity’] = false
}
```

Issue #793, opened 2025-12-18
(DRAFT) Caching Requests


agilpatr:
Need to create a new function in frontend/services/azureFuncs.ts that will cache requests. The post functions can then call this when the user is offline. The key can be the formUrl and the data can be the formData (see [postProvenance, line 84](https://github.com/gosqasorg/asset-provenance-tracking/blob/main/packages/frontend/services/azureFuncs.ts)). Follow procedure in [#806](https://github.com/gosqasorg/asset-provenance-tracking/issues/806) when caching requests and see Vincent’s comments below.

Also write some unit tests for this function, making sure that when the function is called it successfully stores the request inside localStorage (and that we can get the request out of local storage).

Issue #792, opened 2025-12-18
Detect Online/Offline


agilpatr:
To be able to cache a user’s requests while they’re offline, we first need to be able to detect when they’re offline. We can do this by checking if a user is online/offline before making a post request. Create a new function in [azureFuncs.ts](https://github.com/gosqasorg/asset-provenance-tracking/blob/main/packages/frontend/services/azureFuncs.ts) and try to send a small fetch request (ie. ping google.com) and see if the request fails or not (if it fails, then the user is offline). This function can then be called from our post functions to check if the user is offline before attempting to create a record. We’ll handle what to do if a user is offline in another issue, for now you can just return an error or print a message saying they're offline.

For reference, take a look at the fetchUrl function in azureFuncs.ts. This function tries three times to either create or get a record, throwing an error if it fails. The second error (with response == undefined) corresponds to the user being offline.

For this issue, create two tests, under unit tests, to test each case of online and offline. To do this, monkeypatch the native fetch function by defining a function of the same name. Do this in the scope of the test for successfully detecting offlineness, so as not to clobber the true native fetch function itself.

paulErdos:
https://developer.mozilla.org/en-US/docs/Web/API/Window/localStorage

Issue #886, opened 2026-03-11
Disable and Enable Back Online Banner


agilpatr:
When the user clears the records on the offline edits page we want the back online banner to disappear. Given the sensitive nature of this emergency-scenario-oriented feature, we want to ensure that the list of records for which creation was initiated while offline is, once the records are created, only dismissible manually. This is to ensure that the user does not lose access to the record IDs.

Since we've now updated offline mode's layout, we also need to turn on the banners in our new worker (see pseudo code in the [design doc](https://docs.google.com/document/d/1Qdu98p0ZnnkBYjOi8Yb9e5FMOtHBWsVkS8DJ3ywYbDI/edit?tab=t.d1xrpopb77w6#heading=h.bopqf08pobyw)).

agilpatr:
Handled by #901

PR #986, opened 2026-04-24
Implement Detection + Stashing Function and Incorporate into PostProvenance

### LINKED ISSUE #833: Implement Detection + Stashing Function and Incorporate into PostProvenance
1. Define a new function in azureFuncs.ts that takes as input formUrl and formData and returns either 200 if the user is online or 202 accepted if the user is offline. Check if the user is online or not using the detection method from #792. If the user is offline, the function needs to stash the formUrl and formData (using the function from #793), kick off the periodicChecker function (#850) by calling it and not awaiting it, and then return 202.
2. Then call the function in postProvenance right before the fetchUrl call [on line 84](https://github.com/gosqasorg/asset-provenance-tracking/blob/main/packages/frontend/services/azureFuncs.ts#L84). If the function from item one does return 202, throw that status code and an error message (something along the lines of the user is offline but the record has been stashed) from postProvenance instead of the normal return value.
3. We also want to make the offline banner appears when this function is called. This functionality already exists in our detection method, but we'll need to add functionality to our emptyStash function to disable the banner once the user is back online (and display the back online banner).

Notes:
- Implement Banner Issue #861
- [Documentation on status code 202](https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Status/202)

Here is an example of how an async function operates when it is called but not awaited:
Image

HieuNgoGH:
- Created offlineDetectAndStash function that detects whether the user is offline or online using method from issue #792.
- If the user is offline, the formUrl and formData are stashed using function from issue #793. periodicChecker function from issue #850 is also called.
- in postProvenance, this function is called and if the status code is 202, the status code along with error message will be thrown.
- Offline banner is disabled and online banner is enabled in emptyStash function before the return call.

### DIFF
--- a/packages/frontend/services/azureFuncs.ts
+++ b/packages/frontend/services/azureFuncs.ts
@@ -87,6 +87,10 @@ export async function postProvenance(deviceKey: string, record: any, attachments
     
     const fullUrl = baseUrl + "/provenance/" + deviceKey;
     try {
+        // Checks to see if user is offline, stashes record if offline
+        if (await offlineDetectAndStash(fullUrl, formData) === 202) {
+            throw new Error('Status 202: User is offline but the record has been stashed')
+        }
         let response = await fetchUrl(fullUrl, formData);
         return await response.json() as { record: string, attachments?: string[] };
     } catch (error) {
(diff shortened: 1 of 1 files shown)


agilpatr (approved):
I disabled the online banner for now as we don't currently have a way to disable it once enabled, and will add a note to issue #886 to uncomment the line to enable it. Other than that everything looks great!

PR #987, opened 2026-04-26
localStorage Limit and Notify Users When It's Full

### LINKED ISSUE #917: localStorage Limit
We want to find what the localStorage limit is and add error handling for that case (ie. display a message to the user saying that localStorage is full and no more records can be stored).

HieuNgoGH:
Wrapped stashRequest in a try/catch block so when the user hits the localStorage limit of 5MB, a QuotaExceededError will be caught. Once caught, the error is logged and thrown so we can display to the user in a snackbar or another method, that no more records can be stored.

Updated code to include offlineDetectAndStash to catch 'QuotaExceededError' to be caught as well.

### DIFF
--- a/.github/actions/setup-backend/action.yml
+++ b/.github/actions/setup-backend/action.yml
@@ -4,7 +4,7 @@ runs:
   using: "composite"
   steps:
     - name: Setup Node ${{ env.NODE_VERSION }} Environment
-      uses: actions/setup-node@v4.0.4
+      uses: actions/setup-node@v4
       with:
         node-version: ${{ env.NODE_VERSION }}
 
--- a/.github/workflows/be_on_push.yml
+++ b/.github/workflows/be_on_push.yml
@@ -24,7 +24,7 @@ jobs:
     outputs:
       backend: ${{ steps.filter.outputs.backend }}
     steps:
-      - uses: actions/checkout@v4.1.7
+      - uses: actions/checkout@v4
       - uses: dorny/paths-filter@v3
         id: filter
         with:
@@ -39,7 +39,7 @@ jobs:
     runs-on: ubuntu-latest
     steps:
     - name: Checkout
-      uses: actions/checkout@v4.2.0
+      uses: actions/checkout@v4
 
     - uses: ./.github/actions/setup-backend
 
@@ -56,12 +56,12 @@ jobs:
     needs: backend-run-unit-tests
     steps:
     - name: Checkout
-      uses: actions/checkout@v4.2.0
+      uses: actions/checkout@v4
 
     - uses: ./.github/actions/setup-backend
 
     - name: Azure Functions Action
-      uses: Azure/functions-action@v1.5.2
+      uses: Azure/functions-action@v1
       if: github.ref == 'refs/heads/main' && github.event_name == 'push'
       with:
         app-name: ${{ env.AZURE_FUNCTIONAPP_NAME }}
@@ -74,7 +74,7 @@ jobs:
     needs: backend-build-and-deploy
     steps:
     - name: Checkout
-      uses: actions/checkout@v4.2.0
+      uses: actions/checkout@v4
 
     - uses: ./.github/actions/setup-backend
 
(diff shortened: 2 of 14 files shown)


agilpatr (changes requested):
Looks good! Just one thing, the new offlineDetectAndStash function can only throw an error to the snackbar saying that the record has successfully been stashed, but in the case of stashRequest throwing a QuotaExceededError we want it to instead display a message informing the user that their storage limit has been hit and their record could not be stashed. Pull main into this branch and update offlineDetectAndStash to handle the limit reached case and then this issue should be good to go.

HieuNgoGH:
> Looks good! Just one thing, the new offlineDetectAndStash function can only throw an error to the snackbar saying that the record has successfully been stashed, but in the case of stashRequest throwing a QuotaExceededError we want it to instead display a message informing the user that their storage limit has been hit and their record could not be stashed. Pull main into this branch and update offlineDetectAndStash to handle the limit reached case and then this issue should be good to go.

Got it!

agilpatr:
> > Looks good! Just one thing, the new offlineDetectAndStash function can only throw an error to the snackbar saying that the record has successfully been stashed, but in the case of stashRequest throwing a QuotaExceededError we want it to instead display a message informing the user that their storage limit has been hit and their record could not be stashed. Pull main into this branch and update offlineDetectAndStash to handle the limit reached case and then this issue should be good to go.
> 
> Got it!

@HieuNgoGH Awesome, thanks! Sorry for wording this poorly in my first comment, but we'll still need to display to the user that the storage is full. For the QuotaExceededError case we want to return some error code (507 Insufficient Storage might be a good one) in offlineDetectAndStash that we can check for in postProvenance (so we can then throw the error to display to the user). A good way to test is to throw an error from offlineDetectAndStash and see if the browser displays the correct error when creating a record. Let me know if any of this is unclear!

Code in postProvenance to change:
```
if (await offlineDetectAndStash(fullUrl, formData) === 202) {
        throw new Error('Status 202: User is offline but the record has been stashed')
}
// TODO: want a new case here where if the error code === 507 throw error along the lines of
        // "storage limit has been reached, record not stashed"
```

Code in offlineDetectAndStash to change:
```
} catch (error: any) {
        if (error.name === "QuotaExceededError") {
            console.log('Storage limit has been reached: your record has not been stored')
            // TODO: want to return error code 507 (or any error code) for postProvenance to throw to the user
        }
        else {
            // This part is fine to be left as is
            console.log('Error in offlineDetectAndStash: ' + error)
        }
}
```

HieuNgoGH:
> > > Looks good! Just one thing, the new offlineDetectAndStash function can only throw an error to the snackbar saying that the record has successfully been stashed, but in the case of stashRequest throwing a QuotaExceededError we want it to instead display a message informing the user that their storage limit has been hit and their record could not be stashed. Pull main into this branch and update offlineDetectAndStash to handle the limit reached case and then this issue should be good to go.
> > 
> > 
> > Got it!
> 
> @HieuNgoGH Awesome, thanks! Sorry for wording this poorly in my first comment, but we'll still need to display to the user that the storage is full. For the QuotaExceededError case we want to return some error code (507 Insufficient Storage might be a good one) in offlineDetectAndStash that we can check for in postProvenance (so we can then throw the error to display to the user). A good way to test is to throw an error from offlineDetectAndStash and see if the browser displays the correct error when creating a record. Let me know if any of this is unclear!
> 
> Code in postProvenance to change:
> 
> ```
> if (await offlineDetectAndStash(fullUrl, formData) === 202) {
>         throw new Error('Status 202: User is offline but the record has been stashed')
> }
> // TODO: want a new case here where if the error code === 507 throw error along the lines of
>         // "storage limit has been reached, record not stashed"
> ```
> 
> Code in offlineDetectAndStash to change:
> 
> ```
> } catch (error: any) {
>         if (error.name === "QuotaExceededError") {
>             console.log('Storage limit has been reached: your record has not been stored')
>             // TODO: want to return error code 507 (or any error code) for postProvenance to throw to the user
>         }
>         else {
>             // This part is fine to be left as is
>             console.log('Error in offlineDetectAndStash: ' + error)
>         }
> }
> ```

Hey @agilpatr , no worries! Thanks for the extra clear instructions. Here's the update, please feel free to suggest any changes!

agilpatr (approved):
Looks good, thanks for making those changes so quickly!

PR #996, opened 2026-04-28
Adding temp file to trigger deployment attempt

### LINKED ISSUE #995: Reconfigure deployment to bundle package
Updating our node version has pushed our package zip over the deployment size limit, meaning deployments are blocked. This is not a major issue: most of the package zip size comes from node_modules, the directory containing all of the code from all of our dependencies. 

The fact of the matter is that in general, most of the dependency code is not used. For example, many of you are familiar with Python, and probably some of you have used numpy. However, it's essentially never the case that one uses all of numpy, rather, just a few data types and functions. 

The solution to this is to alter our deployment to bundle our code. The process of bundling does something called tree-shaking, where unused dependency code is not bundled. 

This will bring our package size down considerably.

paulErdos:


paulErdos:
Merging to main to trigger Azure Functions Action build step

paulErdos:
Got the error.

`
Offset to Central Directory cannot be held in an Int64.
Error: Failed to deploy web package to App Service.
Error: Execution Exception (state: PublishContent) (step: Invocation)
Error:   When request Azure resource at PublishContent, zipDeploy : Failed to use /home/runner/work/_temp/temp_web_package_5082917173721277.zip as ZipDeploy content
Error:     Package deployment using ZIP Deploy failed. Refer logs for more details.
Error: Deployment Failed!
`

PR #1021, opened 2026-05-11
Added feature flag for Offline Mode

### LINKED ISSUE #1011: Create Offline Feature Flags
We need feature flags for our offline features so that we can disable them on production while they're in development. We'll want flags for any offline features that the user could run into and be confused about (ie. call to [checkOffline in postProvenance](https://github.com/gosqasorg/asset-provenance-tracking/blob/main/packages/frontend/services/azureFuncs.ts#L89), offline banners, etc).

HieuNgoGH:
Added global variable feature flag for Offline Mode to toggle off while in production and on in development we don't confuse our users with our incomplete feature.

### DIFF
--- a/packages/frontend/services/azureFuncs.ts
+++ b/packages/frontend/services/azureFuncs.ts
@@ -15,6 +15,9 @@
 
 import { validateKey } from "~/utils/keyFuncs";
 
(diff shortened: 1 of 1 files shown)


agilpatr (approved):
Looks good!

PR #1028, opened 2026-05-13
Added line to solve local CORS issue when offline

### LINKED ISSUE #1013: CORS Issue Blocking Localhost7071
We ran into an issue where CORS is blocking localhost7071 from localhost3000. In my own testing I wasn't able to recreate the issue, but it was found while trying to create records while offline.

HieuNgoGH:
Because of our PWA feature, the service worker caches all of our pages and thus even when offline, the fetch function calls our frontend url, which is cached so it will return 200, even though we are offline. This line tells the browser not to look into our cache for the request and instead make a real network request instead. This resolves the CORS issue when trying to ping either Google which returns a CORS error or when calling our frontend url, which returns a 200 status due to our pages being cached, which we don't want when offline because it bypasses the important offline code block to enter offline mode.

### DIFF
--- a/packages/frontend/services/azureFuncs.ts
+++ b/packages/frontend/services/azureFuncs.ts
@@ -215,7 +215,7 @@ export async function onlineTestFetch(url?: string): Promise<boolean> {
     }
 
     try {
-        let response = await fetch(url);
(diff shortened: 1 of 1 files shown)


agilpatr (approved):
Nice work!

PR #1039, opened 2026-05-23
added /createRecord API spec

### LINKED ISSUE #964: (DRAFT) Create Spec for CreateRecord

milenapetrov:
Wrote in the required inner fields for provenanceRecord: 'deviceName', 'description', and 'children_key.'

Outer 'attachment' field, supporting either an empty string or an object with file and name.

### DIFF
--- a/packages/frontend/public/openAPI-docs-with-servers.yaml
+++ b/packages/frontend/public/openAPI-docs-with-servers.yaml
@@ -74,7 +74,76 @@ paths:
                   error:
                     type: string
                     example: "Invalid input parameters"
+  /createRecord:
+    post:
+      tags:
+        - GDT API
+      summary: Create a new record
+      description: This endpoint creates a new record and returns a unique record URL.
+      operationId: createRecord
+      requestBody:
+        description: Record creation parameters in postProvenance
+        required: true
+        content:
+          application/json:
+            schema:
+              type: object
+              required:
+                - provenanceRecord
+              properties:
+                provenanceRecord:
+                  type: object
+                  required:
+                    - deviceName
+                    - description
+                    - children_key
+                  properties:
+                    deviceName:
+                      type: string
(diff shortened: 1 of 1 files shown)


milenapetrov:
It seems that children_key and attachments are required for a postProvenance request to succeed. Without them, the request fails on execution. 

Should this still be the case if attachments are optional for the user. 

Does children_key need to be a required value of createRecord in the API if a blank string is hardcoded in the backend for newly created records? (Are there other places in the interface where createRecord/postProvenance is called that require attachments?)

Let me know if you need clarity on these questions.

paulErdos:
Hi @milenapetrov. 

> Does children_key need to be a required value of createRecord in the API if a blank string is hardcoded in the backend for newly created records?

No. The only required fields are Title and Description, just like in the UI. 

One thing I'm noticing: the endpoint. We don't have a /createRecord. Records are typically made using /postProvenance. 

Update: Pardon me, @milenapetrov: we do have a createRecard, here: https://github.com/gosqasorg/asset-provenance-tracking/blob/main/packages/backend/src/functions/httpTrigger.ts

When I tried this, though, it did not seem to use that endpoint.

paulErdos:
Working. 
@milenapetrov, I'd like to offer you the opportunity to demo this at the next Monday meeting. Basically what we're looking for is this: show the docs page, click through the interactive "try it out" (with the network tab open so we can see the request go through), then take the url we get back, and paste it into a new tab and show that it exists. ezpz. 

Screenshot From 2026-06-10 15-49-52
Screenshot From 2026-06-10 15-50-11
Screenshot From 2026-06-10 15-50-18

paulErdos (approved):

milenapetrov:
Hi @paulErdos 

Just saw this, whoops, happy to demo at next week's meeting!

What changed between /createRecord not working and then working on your end?

PR #1043, opened 2026-05-27
Improved emptyStash Function

### LINKED ISSUE #1036: Improve emptyStash Function
During our last meeting we came up with a couple of improvements for the emptyStash function:

1. Handle the case where a stashed record returns empty by throwing an error to be caught by the catch block.
```
// Get the last request stored
let request_name = 'gosqas_offline_stash_' + stash_counter;
// TODO: handle the case where gosqas_offline_stash_ is empty
let request = JSON.parse(localStorage.getItem(request_name) || '{}');
if (request === '{}') {
   throw new Error()
}
```

2. Try to get the record after creating it to confirm it exists, only removing the record from the stash if it returns successfully.
```
// Fulfill the request
const formData = new FormData();
formData.append('provenanceRecord', record);
let response = await fetchUrl(fullUrl, formData)
if (response.status != 200) { throw new Error(`Fetch failed with error code ${response.status}`) }

// TODO: try to get prov, if it's there then it was created
   // if not something failed, leave in stash
```

3. Inside the for loop, right after the try/catch block, call onlineTestFetch() to confirm the user is still online, exiting the loop if they are not.

agilpatr:
Implemented the requested changes for the emptyStash function:
- If the stashed request is empty: delete the request, update the counter, and go to the next request
- Function now tries to get the created record, and if that or creation fails it goes to the catch block
- If in the catch block the user is online, add the record to the failed stash, delete the request, and update the counter (since if the record failed to create and the user is online, then something unrelated to offline went wrong)
- If something went wrong and the user if offline, simply continue to the end of the loop without removing the record
- At the end of every loop checks if the user is still online, exiting the function if they are not
- If a record cannot be formatted to stash in either fulfilled or failed the whole string will be put into the failed stash

I also updated the name of the Offline Creation Page to Offline: Add to Existing Record, as discussed in the meeting.

### DIFF
--- a/packages/frontend/pages/history/offline.vue
+++ b/packages/frontend/pages/history/offline.vue
@@ -35,7 +35,7 @@ their items while offline.
                 <div class="my-4 text-iris fs-1">
                 <p class="text-bold mb-0 device-name">Asset History Records</p>
                 <h1 class="mt-1 mb-1" style="word-break: break-word;">
-                    Offline Creation Page
+                    Offline: Add to Existing Record
                 </h1>
                 </div>
                 <div class="rec">
--- a/packages/frontend/services/azureFuncs.ts
+++ b/packages/frontend/services/azureFuncs.ts
@@ -276,16 +276,21 @@ export async function stashRequest(formUrl: string, formData: FormData) {
 }
 
 async function stashKeysAndRemove(fullUrl: string, stashName: string, request_name: string, stash_counter: number) {
-    let keys = [];
-    let currentKey = fullUrl.split("/")[fullUrl.split("/").length - 1];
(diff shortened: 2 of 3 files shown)


agilpatr:
**Note**: Since fetchUrl doesn't return a status code upon failure, emptyStash() checks online status after the record fails to create. There is a niche case here that the record fails offline, and then the user comes back online just after, falsely putting the record into the failed stash. We plan to allow users to retry creation anyways, so this shouldn't be a big issue, but in the future it might be worth modifying fetchUrl to throw codes that other functions can check.

paulErdos (approved):
Very good testing, Alison!

PR #1048, opened 2026-05-29
Redirect When Creating Records Offline

### LINKED ISSUE #1033: Redirect User After Creating A Record While Offline
When the user creates a record while offline we'd like the page to redirect to the Record page (/history/offline) and auto fill the data using the newly created key and any information for that key that exists in the stash.

This should occur for records created from all three creation pages (CreateDevice.vue, CreateContainer.vue, and CreateRecord.vue).

agilpatr:
If you create a record from any of the three creation pages while offline you will now be redirected to the offline history page. When you're redirected the key will autofill (since redirects only happen on record creation only the key is sent, since the rest of the information is stashed to be created once the user comes back online).

To make vue's router work offline we needed to preload the offline history page, which has the added benefit of making sure the page will load offline when redirected to (even outside the PWA).

Screenshot 2026-05-29 at 3 38 31 PM

### DIFF
--- a/packages/frontend/components/Forms/CreateContainer.vue
+++ b/packages/frontend/components/Forms/CreateContainer.vue
@@ -388,8 +388,7 @@ export default {
             } catch (error) {
                 // If the user is offline navigate to the offline history page instead
                 if (!(await onlineTestFetch())) {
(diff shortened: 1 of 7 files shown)


paulErdos (approved):
Working

PR #1049, opened 2026-05-31
Periodic Checker Falsely Skipping

### LINKED ISSUE #1042: Periodic Checker Falsely Skipping
While testing I found a bug with the periodicChecker. If you create a record while offline and then close the tab while the periodicChecker is running, then the localStorage item 'gdt-awaiting-conectivity' will stay set as true when it should go back to false. This creates the issue that when you create another record while offline it will skip the periodicChecker as it thinks it's already running when it isn't.

We want to modify periodicChecker so that it only can run one instance of itself, _and_ will only be listed as running when it actually is. Whether we want to keep using 'gdt-awaiting-conectivity' in localStorage or use a different method to track this is up to the developer.

HieuNgoGH:
- Added line to fix bug where when periodicChecker is running and the tab is closed. 'gdt-awaiting-connectivity' will now start as 'false' when a new tab is opened so that it can when another record is created periodicChecker won't think it's already running due to the if-clause in its function.

- Changed 'gdt-awaiting-connectivity' to false after the stash is emptied periodicChecker upon online detection and a successful onlineTestFetch call when the user is back online. 

-  Added extra n to 'gdt-awaiting-conectivity' in azureFuncs.spect.ts to align and pass the tests.

### DIFF
--- a/packages/frontend/services/azureFuncs.ts
+++ b/packages/frontend/services/azureFuncs.ts
@@ -16,7 +16,7 @@
 import { validateKey } from "~/utils/keyFuncs";
 
 // Feature flag to turn ON/OFF Offline Mode features while in development
-export var offlineModeFeatureFlag = true;
+export var offlineModeFeatureFlag = false;
 
 // Global variable used to control the display of offline banner on create pages
 export var displayOfflineBanner = false;
@@ -25,13 +25,6 @@ export var displayOfflineBanner = false;
 export var displayOnlineBanner = false;
 
 localStorage.setItem('gdt-awaiting-connectivity', 'false')
-
-console.log("AT START")
-console.log("STASH_COUNTER: " + localStorage.getItem('stash_counter'))
-console.log("STASH_COUNTER NUMBER TYPE: " + typeof(localStorage.getItem('stash_counter')))
-window.addEventListener("load", () => {
-    console.log("LOADED ON PAGE LOAD")
-})
 
 // method takes the base58 encoded device key
 export async function getProvenance(deviceKey: string) {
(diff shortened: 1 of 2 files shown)


agilpatr:
Discussed in the offline meeting, notes below and in planning doc:
Some sort of session id to track if a tab is already open
- Create session ID when periodicChecker starts and set gdt-awaiting-connectivity to true
- On tab closing delete session ID (https://www.geeksforgeeks.org/javascript/how-to-detect-browser-or-tab-closing-in-javascript/)
- Opening new tab: if gdt-awaiting-connectivity is true and session ID does not exist re-enter periodicChecker (which will create new session ID)

HieuNgoGH:
Hey @agilpatr!, upon researching on session ID to implement in this issue. Being that session ID needs s network connection to reach the backend, it will not work with our PWA in offline mode. We could use a local token, but that also expires when the PWA is closed and it would need another network call to get another token. Please do advise if you have any other ideas!

HieuNgoGH:
@agilpatr, diving further into the issue, since PWA's use a single dedicated window, there would be no additional tabs. It also being in the middle-ground of an app off the app store and a website, it's very unlikely someone would have two instances of the same app running simultaneously.

agilpatr (changes requested):
Hey @HieuNgoGH, thanks for the update! Good to know that this won't be an issue with the PWA, though for devs and to prevent problems down the line I think it's still worth finding a solution that prevents records from getting stuck in the stash. I think a good middle ground would be having 'gdt-awaiting-connectivity' set to false when a new tab is opened (like you have now) and then also checking if the stash_counter is 0. If it's not 0 then there's still items in the stash to create, so wait 5-10 seconds (to make sure all previous instances of the periodicChecker have stopped) and then call the periodicChecker again. For this to work we'd have to modify our while loop in periodicChecker to also break if 'gdt-awaiting-connectivity' is false (would probably need the same check in the for loop in connectivityChecker and emptyStash). That way we know all requests will be fulfilled even if a tab is closed in the middle of the process, while also hopefully only running one instance.

Let me know if this solution makes sense or if you have another idea, and feel free to make any needed changes to it!

HieuNgoGH:
Closing until confirmation on what to do next with offline mode.

PR #1051, opened 2026-06-02
Allowed Online/Offline Toggling and Created Record While Offline

### LINKED ISSUE #903: Create Record While Offline
We want to try creating a record while offline to make sure that the record is successfully stored, then go back online to make sure the record is created. We should use the same path/method as regular record creation (ie. call postProvenance which should cache the request when offline and start the periodicChecker), which will likely require us to simulate offline temporarily. We'll want to write this test in the [azureFuncs.spec.ts](https://github.com/gosqasorg/asset-provenance-tracking/blob/main/packages/frontend/test/unitTests/azureFuncs.spec.ts) file.

### LINKED ISSUE #1045: Allow Toggling For Online/Offline
For testing we'll need to alter the code to allow us to toggle on/off online at will. For this we'll need to make two changes:

The first is to alter onlineTestFetch to reference a global variable in azureFuncs.ts for it's standard url. This will allow us to dynamically alter the value of this url during testing, allowing us to modulate whether the code thinks it's online or not.

The second is to alter the way we stash requests. Presently we stash the entire url. If we instead stash only the url past the baseUrl, we can then fill in the baseUrl using a global variable when fulfilling requests. We can in testing dynamically set request to fail when needed.

agilpatr:
Updated our offline code so that in testing we could more easily simulate online/offline:
- onlineTestFetch now references a global variable that can be set from the test file
- stashRequest now stores the key rather than the full url (previous functions and tests have been updated as well). Only storing the key since dev/prod store /provenance/ in their baseUrl while localhost does not.

Also wrote an integration test for creating a record while offline (to confirm that the above changes did allow us to test online/offline). Normal fetch requests fail from frontend test files (formData gets converted to plain text when sent) so I've mocked it to post with form url encoded data instead.


paulErdos (approved):
Excellent work on the testing with this, @agilpatr

PR #1060, opened 2026-06-15
859 email validation

### LINKED ISSUE #859: Email Validation
This is a fullstack task. While we're waiting for UI design, putting a text box somewhere is fine. We can make things look nice later. 

The purpose of this task is to prevent folks from signing up emails they do not own: bogus emails fill our database with junk, other people's emails can turn us into a spam-cannon, all the while costing us funds. 

We are committed to having no notion of users or accounts. To do this, we'll do two things when a user hands us an email:
1. We'll hand it to the backend. No matter whether we have it already or not, we'll send to the email a six-digit code. 
2. In the UI, we'll toggle visibility to a text box right next to the email text field, where the user can enter the code, and press a button. 
3. When the button is pressed, the code is transmitted to the backend. 

If the codes match, we store the email under the tags the user wishes to access.

mishwilso:
Adds the email validation flow for subscriptions. Users can now enter their email, receive a verification email with both a 6-digit code and an auto-verify link, and complete verification either by clicking the link or typing the code manually.


packages/backend/src/functions/httpTrigger.ts
```diff
@@ -454,7 +454,7 @@ async function notifySubscribers(deviceKey: string, context: InvocationContext):
         return;
     }
 
-    const from_address: string = "DoNotReply@8577d69b-9011-4385-abec-cfe9325dbfe6.azurecomm.net";
+    const from_address: string = "DoNotReply@091bd21c-5093-45ed-9479-ad92fef9d66e.azurecomm.net";
```
paulErdos: Hey @mishwilso! Just want to check on why this changed. If this is a dev/prod thing, let me know and I'll put them both into an Azure env var and give you its name. Otherwise, let me know what motivated the change. Thanks!

paulErdos:
Hey @mishwilso! I'm seeing something unusual: the EmailNotificationSignup.vue file, and it appears the element itself, was deleted, and the element source wasn't, for example, moved to another file?

Possibly an artifact of merging main?

mishwilso:
Hi @paulErdos!

Sorry for the late reply. EmailNotificationSignup.vue was made by another developer. I think they made it for testing their code with the flow of email verification. I think I resolved the issue of the element source not being deleted- it was referenced in record/[devicekey].vue

paulErdos:
> Hi @paulErdos!
> 
> Sorry for the late reply. EmailNotificationSignup.vue was made by another developer. I think they made it for testing their code with the flow of email verification. I think I resolved the issue of the element source not being deleted- it was referenced in record/[devicekey].vue

Thanks for the (estimated) provenance history on that element, @mishwilso!

paulErdos (approved):

packages/backend/src/functions/httpTrigger.ts
paulErdos: Note for devs of the future: dev's personal azure account sender email. After a bit of debugging, set back to our production email in another more recent branch.

PR #1074, opened 2026-06-18
Created Syncing/Failed Stashes

### LINKED ISSUE #1040: Create Stash for Syncing and Failed Records
To match our FIGMA design we'll need to keep track of records that are syncing and the records that have failed to create.

For syncing records, update emptyStash() to move the record to a new syncing stash after getting it from the gdt_offline_stash. Then once the record has been created or failed to create remove it from the syncing stash.
The flow should be something like this:
- get record from offline_stash -> store in syncing_stash and remove from offline_stash -> create the record -> remove from syncing_stash and store in fulfilled_stash

The failed stash already exists, we just need to make sure records added to the failed stash are also removed from the syncing stash.

We'll want to update our offline edits page to display both of these stashes and implement the "retry syncing" and "edit submission" button for failed records. On top of that we will also want to update the "dismiss all edits" button to also clear the failed and syncing stashes.

agilpatr:
Created stashes for both syncing and failed records, displayed the keys on the Offline Edits page, and implemented the "retry syncing" and the "edit submission" buttons. Created functions stashOfflineRequest and removeOfflineRequest to store/remove a specified record from the syncing, fulfilled, or failed stash (queue stash is formatted differently so it couldn't be part of these functions).

Retry Syncing will simply try posting the record again, refreshing the page and displaying the key as fulfilled if it succeeds and displaying a snackbar error if it fails (will remain in failed stash if retry fails).

Edit Submission will bring users back to either the Create Record or Create Group page to edit their submission. Records created from these pages will either be added to the fulfilled stash upon success or left in the failed stash (which doesn't allow duplicate keys).

**Note**: Offline mode's stash layout is likely to change soon, however this PR was already close to completion by then and fully implements the offline edits page. There will likely need to be changes once we update the layout, but parts of this should still be helpful.


paulErdos:
Hey @agilpatr! I wonder if you've got time to stop by an office hour this week. Some of the work in this PR is unclear, and I think it might be fastest to chat through it on a call. 

We can also divvy up the merge conflicts and get them knocked out.

agilpatr:
Hey @paulErdos, I should be able to drop by one of the office hours today! Otherwise I can drop by sometime next week.

paulErdos:
Hey @agilpatr! I'm seeing a conditional that's no needed in [packages/frontend/components/Buttons/LargeToggle.vue](https://github.com/gosqasorg/asset-provenance-tracking/pull/1074/files/0d291a5795649d6b1d6196c38a7bf5a278f88cd7#diff-91f52dd59bdd61964e9073828abde8c03e04d84a47e7bc90c70fdcb4e0785365) on line 35.

packages/frontend/components/Buttons/LargeToggle.vue
```diff
@@ -21,6 +21,29 @@
...
+        rightLabelStart: false
+      }
+    },
+    mounted() {
+      // If we're redirecting from offline-edits and the record is a group, go to the group tab on load
+      let isGroup = sessionStorage.getItem("gdt-redirect-isGroup");
+      const previousUrl = window.history.state.back;
+      
+      if (isGroup && previousUrl === "/offline-edits") {
+        if (isGroup === "true") {
```
agilpatr: this.rightLabelStart = isGroup === "true";

paulErdos: This conditional is not needed: the boolean expression in the outer conditional is a conjunction. We can only get here if both expressions on either side of the conjunction are true. This needs to be cleaned up before I can approve.

packages/frontend/pages/offline-edits.vue
```diff
@@ -179,7 +208,112 @@ methods: {
...
+            for (let i = 0; i < failedRequests.length; i++) {
+                let fullUrl = failedRequests[i][0][1];
+                let requestKey = fullUrl.split("/")[fullUrl.split("/").length - 1];
+                // If we find the request store it and exit the loop
+                if (requestKey == key) {
+                    stashedRecord = JSON.parse(failedRequests[i][1][1]);
+                    break
+                }
+            }
+
```
paulErdos, line 218: Add a conditional at line 227 to check whether stashedRecord is undefined, in which case, we can display a custom error message to the user, and then return from execution, rather than falling into the catch block, that would display in the snackbar to the user a confusing message of the form "cannot read properties of undefined (reading 'children_name')".

packages/frontend/pages/gdt.vue
```diff
@@ -87,8 +87,18 @@ export default {
         }
     },
     async mounted() {
+        // If we're redirecting from offline-edits and the record is a group, go to the group form on load
+        let isGroup = sessionStorage.getItem("gdt-redirect-isGroup");
+
+        if (isGroup === "true") {
```
paulErdos: Here's another instance of that way of using boolean expressions to set values without using a conditional. Opportunity to save some lines.

packages/frontend/pages/history/offline.vue
```diff
@@ -240,9 +253,13 @@ methods: {
```
paulErdos, line 250: Add a comment to note that if we're offline, ie, the fetchUrl in postprov throws an error, we won't reach any of the following code.

paulErdos:
Items noted during office hour code review:
* Dismiss all is not catching the underscore-named items in the stash
* When creating a new record or adding to an existing record, let's have the snackbar message show up as green, as we're letting the user know that their record has been recorded / stashed. 
* Go ahead and create a new ticket for this: When creating a record offline, we're taken to the offline history page (intended). However, the offline history page itself does not mirror the functionality of the non-offline history page in the following way: the record history elements are not shown, and the user cannot add new entries, they can only alter the title and description. 
* Separate ticket from the one this PR references, although probably best to have it in the same ticket as the separate one from the above bullet point: We're talking about altering the concept of the offline history (/history/offline?key=the_key) page to be, rather than a separate page, literally the same page, as the history page does (in essence) two things: grabs information and displays it, and adds information and stores it. For offline, we could handle this in one of two ways: either have getProvenance notice whether we're offline, and choose whether to invoke the api (online) or retrieve information from the stash (offline), or we could have two getProvenance style functions, ie, the normal one, and a new one that retrieves from stash.

packages/frontend/components/Provenance/CreateRecord.vue
paulErdos: The code staring line 305 in the conditional that checks if records came back null can't be hit with this redirect in place. Of course it made sense to put it here initially. With the stashes you've written, it makes sense to drop this, as we want to add items as indeed you're doing in the line305 conditional block. To get that to be hit, try dropping this line-300 redirect.

paulErdos:
Plan to continue the review in office hours after the Monday meeting. All reviewed aside from:
* CreateContainer,
* CreateDevice, and
* LargeToggle

Update: done

paulErdos:
More items noted during add'l office hour code review:
* To sidestep the error induced by hasParent being a function, while still needing to alter the hasParent data item in group child records, swap from renaming the hasParent variable name to instead renaming the function, both where it is defined, and in all instances where it's called, as this requires fewer renames, meaning it's easier to ensure no bugs are added by a bulk-rename operation. (See createDevice line 136 in this branch)
* Try setting stashedRecord (see CreateRecord line 146 and 255) in the data function on line 129, as if this works, we'll save the user a few function calls' worth of execution time. 
* When creating a record offline, we're double-snackbar-ing the user (the snackbar message shows up twice)
* Go ahead and create a separate ticket for this: For fulfilled records, we'd like to allow users to make edits if they go back offline. One way to approach this would be to keep the entire provenance json as a value in the fulfilled stash.

paulErdos:
One note: with all the comments and items we're seeing for edits, some are related to this ticket, and some are not strictly related. As needed, go ahead and create separate tickets.

paulErdos:
Go ahead and create a separate ticket for this: Add to design doc, a sub-page that lists everything we're doing with localStorage. We have several caches, and we have stored state as a few variables, and it will save us time down the road to be able to have a spec to reference (spec here distinct from the api spec, rather, a localstorage-usage spec sheet)

(and in the ticket, go ahead and add a link to the google doc)

agilpatr:
Hi @paulErdos, I made the requested changes:
- Removed unneeded conditional in LargeToggle.vue and gdt.vue
- Now displays custom error on the offline edits page if stashedRecord is undefined
- Dismiss All button now removes records with ’_’ and ’-’
- Added comment on the offline page noting that code after postProvenance is not reached if the record fails to create
- Updated the 202 snackbar to be green on all pages
- Removed redirect to the offline history page from the regular history page, and moved error message snackbar to the catch block (rather than the unreachable if statement)
- Set stashedRecord in the data() section to save function calls
- Removed hasParent variable rename in CreateDevice and instead renamed the hasParent function and all of its invocations
- Stacked snackbars when creating multiple records offline

I’m working on making the new issues mentioned in this PR now, and they should all be created in the next day or two.

paulErdos (approved):
Reviewed in office hours.

paulErdos:
Thanks @agilpatr! I went ahead and resolved two minor merge conflicts for ya. Glad this one is finished!

PR #1084, opened 2026-06-23
Updated Offline Feature Flag

### LINKED ISSUE #1083: Update Offline Feature Flag
A couple offline features are not being disabled by our feature flag, including:
- offline banners, which mention incomplete features
- redirect to offline history page after creating a record offline

We'll want to update our feature flag to disable the above features when set to false.

agilpatr:
Updated the offline feature flag to disable banners and offline redirects. Also updated the flag to be modifiable from other files to avoid merge conflicts with PR #1051.


paulErdos (approved):
Having done a bunch of feature flagging myself this eve, I know how tedious this can be. 

One idea for setting the status of the feature flag: check the value of the backend url. If it's not prod (ie, if it doesn't include 'local' or 'gosqasbe'), then we know we're not on prod and are clear to have offline mode un-hidden. 

Up to you if you'd like to set it in said way. It does set aside having to manually set the flag. If you like the idea, go ahead and create a new issue and send me a PR for just that. However if you prefer it this way, no worries.

paulErdos (approved):

paulErdos (approved):

paulErdos (approved):
Funny that GitHub allows multiple consecutive PR approvals.

PR #1088, opened 2026-06-26
Disable "Add To Group" UI element for records that already have a parent

### LINKED ISSUE #1063: Disable "Add To Group" UI element for records that already have a parent
Provenances for such records have a "has parent" boolean:

```[{"record":{"blobType":"deviceInitializer","deviceName":"foo #1","description":"bar","tags":[],"children_key":"","hasParent":true,"isReportingKey":false},"attachments":[],"deviceID":"091c4399f377cab1d70bb58f5c0c88e7bd121a2da0f7293811e9b0dc9c8b4321","timestamp":1781571610024}]```

HieuNgoGH:
Added code to only display UI element 'Add to Group' when a record doesn't have a parent.

Parent Key:
parent

Child Key:
child


paulErdos:
Okay great! This solves the issue for records created during group creation. When adding a record to an existing group, the box still pops up. I'll create another issue for that. This on its own is solid. Thanks, @HieuNgoGH!

paulErdos (approved):

PR #1089, opened 2026-06-29
Tag Specific Email Signups

### LINKED ISSUE #1054: Tag-specific signup and notification frontend
Implement the frontend based on Nina's design [UI Design: Empower users to sign up on a tag-specific basis
 #827](https://github.com/gosqasorg/asset-provenance-tracking/issues/827) 

Display http status to user in the snackbar.

agilpatr:
Implemented Nina's designs to sign up for tag-based email notifications on the Create Device, Create Container, and Create Record pages. Updated our TagInput component to be able to hide the Suggested Tags field and show multiple components on the same page. Note: two of the TagInput functions were moved to a functions file, this is because if they were global they overrode each other and if they were in export default the eventListener could not access them.

Since the tag specific email signup function doesn't exist yet I can't have it display an http status. Currently the tags and email are stored and then discarded upon creation (these values are stored in variables on the create pages and can be used similarly to how the regular email signup used to work).

This is what the new field looks like in both light and dark mode (note: only have the tag specific email signup checkbox to match current changes to main):
Screenshot 2026-06-29 at 2 39 13 PM
Screenshot 2026-06-29 at 2 39 22 PM


paulErdos:
Hi @alison, is this presently feature-flagged?

agilpatr:
Hi @paulErdos, it is not currently feature-flagged but I can create a flag if you'd like.

paulErdos:
> Hi @paulErdos, it is not currently feature-flagged but I can create a flag if you'd like.

Hi @agilpatr, thanks, I'm working on flagging by env var so the feature appears only on dev; I'll take care of it for this one. Thanks for leaving it un-flagged for review.

paulErdos (approved):

PR #1097, opened 2026-06-30
Created Spec for upgradeProvenance

### LINKED ISSUE #972: (DRAFT) Create Spec for upgradeProvenance

agilpatr:
Created spec for the upgradeProvenance endpoint. This endpoint takes a legacy record and converts it to a provenance record. There are multiple possible responses with status 200 so I listed all of them as examples, see the different cases below:
- Key Successfully Upgraded: Legacy record exists and was converted to a provenance record
- Key Already Upgraded: Provided key is already a provenance record
- Key Doesn't Exist: Key does not exist as a legacy or provenance record
- Invalid Legacy JSON: Legacy key exists but the record format is invalid (returns status 404 but doesn't throw it so actual status code is still 200)

Then if the key format is invalid or the backend cannot be reached we get error code 500.

### DIFF
--- a/packages/frontend/public/GDT-OpenAPI-Spec.yaml
+++ b/packages/frontend/public/GDT-OpenAPI-Spec.yaml
@@ -195,3 +195,49 @@ paths:
                   error:
                     type: string
                     example: "Invalid input parameters"
+  
+  /upgrade/{deviceKey}:
+    get:
+      tags:
+        - GDT API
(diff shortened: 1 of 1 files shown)


paulErdos:
Solid. Operating as intended. Thanks, @agilpatr!

paulErdos (approved):

PR #1100, opened 2026-06-30
Added Tags and Reporting Key to Group Creation Spec

### LINKED ISSUE #966: (DRAFT) Incorporate Into Group Creation Spec: Tags and Reporting Key

agilpatr:
Added tags and a reporting key to the group creation spec and updated the example status 200 group to have tags and a reporting key.

Note: I did have to upgrade target from ES2020 to ES2022 in backend/tsconfig.json, otherwise [line 1364](https://github.com/gosqasorg/asset-provenance-tracking/blob/main/packages/backend/src/functions/httpTrigger.ts#L1364) in createGroup would fail (since .at() was added in 2022).


paulErdos (approved):
Great work, @agilpatr!

PR #1107, opened 2026-07-06
added fetch retry utility function + updated createGroup function

### LINKED ISSUE #1092: Create fetch-retry function in httpTrigger + improve createGroup
This ticket has two parts. The first is to create a utility function. The second is to use that function to improve code quality in /createGroup.

Part 1:
Networks are fallible. In the frontend, we have a handy function fetchUrl. This adds retries for responses that come back with 200 < status code <= 300. 

We'd like an equivalent of this in the frontend. 

Part 2:
Presently, createChildren does a manual retry. We'd like to move this retry to createChild, and do this by calling the fetch-retry function createChild.

deneille:
The following additions happened:

- the `fetchWithRetry` function was created for the backend which is similar to the `fetchUrl` function on the frontend. It retries a max 3 times until it gets a responses within the 2xx range.
- the fetch is replaced in the `createChild` function and uses the new `fetchWithRetry` function. The for loop within the `createChildren` function is changed to no longer look for retries but it is a for loop that loop through the custom titles instead.


paulErdos (approved):

paulErdos (dismissed):
Pardon, approved too hastily: would you please resolve the merge conflicts? Thanks.

deneille:
Conflicts should be resolved now.

agilpatr (approved):

PR #1108, opened 2026-07-06
Continue "Disable 'Add To Group' UI element for records that already have a parent"

### LINKED ISSUE #1098: Continue "Disable 'Add To Group' UI element for records that already have a parent"
Major progress so far. Records created during group creation now lack the "add to group" field. See notes 1 and 2.

However, records created independently, that are then added to a group, do not trigger the conditional rendering. 

The reason for this is that when a record is added to a group, the toplevel provenance data is not changed. This is a bug.

Instead, a record history element is added with "hasParent" set to true, while the toplevel record history is still set to false.

This ticket requires two alteration domains:
1. Frontend json building, and
2. Backend API json building (intended to mirror the frontend)

Here is a view of exactly what is going on:
Image

Notes:
1. https://github.com/gosqasorg/asset-provenance-tracking/pull/1088
2. https://github.com/gosqasorg/asset-provenance-tracking/issues/1063

HieuNgoGH:
Implemented a much simpler solution with the use of a shared reactive variable.

The 'Add to Group' field is no longer present in an independently created record's history page.
indiv_record

### DIFF
--- a/packages/backend/src/functions/httpTrigger.ts
+++ b/packages/backend/src/functions/httpTrigger.ts
@@ -3,7 +3,7 @@ import JSON5 from 'json5';
 import * as z from "zod";
 import { webcrypto as crypto } from 'node:crypto';
 import { TableClient, AzureNamedKeyCredential } from '@azure/data-tables'
-import { app, HttpRequest, HttpResponseInit, InvocationContext, InvocationHookContext } from "@azure/functions";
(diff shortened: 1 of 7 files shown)


HieuNgoGH:
Hey @paulErdos!, this has been fixed and working!

agilpatr (approved):
Looks good, thanks!

PR #1110, opened 2026-07-06
Remove Manual Setting for Offline Feature Flag

### LINKED ISSUE #1109: Automatically Set Offline Feature Flag
Based on Vincent's comments on PR #1084, update the offline feature flag to look at the backend url and automatically turn off offline features if we're on production.

agilpatr:
Updated the offline feature flag to be disabled by default, but if we're on dev (ie. localhost or dev site) then enable the offline features.


paulErdos:
Nice use of Nuxt features defining the feature flag in one place and letting auto-import handle putting it where it needs to be elsewhere.

paulErdos (approved):

PR #1122, opened 2026-07-09
1076 apiv2 backend development spec addentry

### LINKED ISSUE #1076: APIv2 Backend Development + Spec: AddEntry
We want to empower users to programmatically update records (incl. group parents, which are just records) using the API. 

eg,
Given a record key, hand the backend some json (refer to API reverse engineering doc in the engineering team docs folder --> create a record and add an entry, and observe the network interaction and take a look at the json), and have the record itself receive the update.

There are two ways to do this:
1. Create a wrapper for postProvenance. Presently a user _could_ navigate interacting with postprov in its particular way to add an update, but we don't want to force users to do that (that's an APIv1 thing). Instead, we can give them a simple /addEntry endpoint that takes just what they need to add (eg, description, tags, attachments), and then just call postProv from that handler function.
2. Create an entirely new function using code from the apiv1 update tests. See [1]. 

Out of the two, the former is cleaner than the latter. 

What's needed for this ticket are the following:
* The new /addEntry endpoint and handler function
* A test
* The spec

Note that the spec itself is a really easy way to do testing in development. 

Regarding the test:
The test will be very simple:
1. fetch to /createRecord so the test has a record to update
2. fetch to the new /addEntry (note: check for 200 vs not-200)
3. fetch to getProv, to see if the entry text landed

Notes:
1. https://github.com/gosqasorg/asset-provenance-tracking/blob/main/packages/backend/test/IntegrationTests/Live/v1/update.test.ts

JaraR:
- Added a new /addEntry endpoint and handler function that can take record entry descriptions, tags, attachments, and annotate requests and posts to specified records.
- Added integration tests to cover the above.
- Added an OpenAPI spec to cover the /addEntry endpoint.


paulErdos (approved):
Great work, @JaraR

PR #1129, opened 2026-07-09
Separated Subscribe and Unsubscribe Features

### LINKED ISSUE #1094: Separate unsubscribe and subscribe operations in emailNotificationUtils > updateNotifications
Presently updateNotifications is suffering from divided attention.

We'd like to split this function into three:
* One for the overhead in lines 57-67
* One for unsubscribing
* One for subscribing

Note the boolean flag in the function signature. We'd like to drop this by separating the function's operations.

agilpatr:
Separated updateSubscriptions into three functions:
- setupBlobClient: overhead from lines 57-67, defines the blob name and client
- unsubscribeFromNotifications: unsubscribes a user from notifications
- subscribeToNotifications: subscribes a user to notifications (note: id generation used to be in a loop [to handle the case that previous emails didn’t have ids yet], however during a code review Rob mentioned this wasn’t strictly necessary and added complexity to the code, so I’ve removed it here. Now an id will only be added for the new email being added)


paulErdos:
Adding notes we've discussed during the review:
* Functionalize the block blob upload in subscribe and unsubscribe, then call in both. If the message can be moved to the catch block, the entire body of the try block can be made into a function and called in each instance, saving ~20 lines. 
* Functionalize the emails / emailIDs list defs, and if(exists) { ... if(catch) { ... } } blocks in both subscribe and unsubscribe into a function that is called in each, passed blob params, and returns the lists.

agilpatr:
Hey @paulErdos, I made the changes that we discussed during the code review meeting. Let me know if there's anything else I should change!

paulErdos (approved):

PR #1138, opened 2026-07-13
Drop group creation attribute auto-inheritance

### LINKED ISSUE #1091: Group Creation Web UI: Drop Attribute Auto-Inheritance
Presently, our group creation frontend automatically annotates attachments, tags, and the group description. After a usability meeting with Victoria and Rob, we'd like to drop these operations. 

To understand the logic, consider the following: the torniquets inside a box of torniquets shouldn't be described as "box of torniquets".

CAI-yw:
### Summary
Group creation was copying the group's description, tags, and attachment onto every child and the public key record. 

This PR removes that copying in` CreateContainer.vue`. Since the public key also gets pushed into `childrenDeviceList`, I dropped the auto-inheritance for it too for consistency. Let me know if we'd rather keep it for the public key.
Screenshot 2026-07-13 at 4 20 39 PM

### Changes
Now children and the public key get a blank description, no tags, and no attachment. The group's own record is untouched. 
The public key keeps its `publickey` tag, since that's its own type marker, not inherited group metadata.

Out of scope: the backend's v2 createGroup endpoint — the live UI doesn't call it, so it wasn't touched.

### Tests

**1. Create a group with a description, tags, and an attached image, plus 2 children. Confirm the group record shows all three, and each child record shows none of them.**
Screenshot 2026-07-13 at 3 06 05 PM
Screenshot 2026-07-13 at 3 06 26 PM

-----
**2. Repeat Test 1 with `"Create Public Key"` checked. Confirm the public key record has no inherited description/tags/attachment, but still carries the `publickey` tag and renders correctly as "Public Key: ..." on its record page.**
Screenshot 2026-07-13 at 4 41 21 PM

publickey

-----
**3. Create a group with _custom child titles_. Confirm titles are still applied correctly and children still have no inherited tags, attachments, description.**
Screenshot 2026-07-13 at 4 43 49 PM
Screenshot 2026-07-13 at 4 44 20 PM
Screenshot 2026-07-13 at 4 44 37 PM

### DIFF
--- a/packages/frontend/components/Forms/CreateContainer.vue
+++ b/packages/frontend/components/Forms/CreateContainer.vue
@@ -303,12 +303,12 @@ export default {
                         await postProvenance(childKey, {
                             blobType: 'deviceInitializer',
                             deviceName: childName,
-                            description: this.description,  // need to see if we want a special description when making a child
-                            tags:this.tags,
+                            description: '',
+                            tags: [],
                             children_key: '',
                             hasParent: true,
                             isPublicKey: false
-                        }, this.pictures || [])
+                        }, [])
                         
                         childrenDeviceList.push(childKey);
                         childrenDeviceName.push(childName);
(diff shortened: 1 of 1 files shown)


agilpatr (approved):
Looks great, nice work giving examples in the PR!

PR #1139, opened 2026-07-14
1121 bug repeated email verified dialog pop up

### LINKED ISSUE #1121: Bug: Repeated "Email Verified" Dialog Pop-Up
When emails are verified using the link in the validation email, the "email verified" popup appears with each new record history element added. 

Note: before this issue can be addressed, the link in the validation email will need to be uncommented, and the changes deployed to dev. Reach out to Vincent, Alison, or Jara to deploy.

deneille:
Issue: `autoVerify` in `EmailNotification` modal never clears the token + code after email is verified and always re-mounted on history record page.

Solution: This uses `emit` to tell the `history/[deviceKey]` page to clear the `token` and `code` after it auto verifies the email. 

Now when you create a record and verify via email, the email verification modal appears only once after clicking the email verification link.


agilpatr (approved):
Nice work!

PR #1140, opened 2026-07-14
Fixed 500 Error on Invalid API Request

agilpatr:
We were getting status code 500 when fetching the url "https://gosqasbe.azurewebsites.net/api/provenance/totallygarbage". This is because the path is technically correct (since in provenance/{deviceKey} to reach the endpoint deviceKey just needs to have a value) and when decodeKey threw a custom error without a status code it defaulted to 500.

I've updated the code to throw status 400 (bad request, since the endpoint is correct it's just the key that's invalid) when decodeKey gets an invalid key. I've done the same for postProvenance as well (since it has the same {deviceKey} issue). **Note**: If we do decide we want this to throw 404 instead of 400 just message me either here or on Discord and I can update it.

**Note: Leaving this in draft until #1135 is merged (since the getProvenance and postProvenance specs from that PR will have to be updated to match the new 400 error code).**

Below is what the page now looks like when the key is invalid.
Screenshot 2026-07-13 at 8 24 57 PM

### DIFF
--- a/packages/backend/src/functions/httpTrigger.ts
+++ b/packages/backend/src/functions/httpTrigger.ts
@@ -268,6 +268,7 @@ async function pathExists(containerClient: ContainerClient, path: string) {
     }
 }
 
+// Check if can place status check in here
 async function convertLegacyProvenance(containerClient: ContainerClient, key: string | Uint8Array) {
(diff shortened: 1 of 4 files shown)


agilpatr:
Created PR on the wrong branch, so I'm closing this one and opening a PR for the correct branch.

PR #1141, opened 2026-07-14
Fixed 500 Error on Invalid API Request

### LINKED ISSUE #953: 500 error on invalid API request
Attempting to make a request from an invalid/non-existing API endpoint will return a 500 Internal Server Error.

Typing the following into node:
```
let foo = await fetch("https://gosqasbe.azurewebsites.net/api/provenance/totallygarbage")
foo
```

we return this as a response:
```
Response {
  status: 500,
  statusText: 'Internal Server Error',
  headers: Headers {
    'content-length': '0',
    date: 'Thu, 09 Apr 2026 19:47:48 GMT',
    'request-context': 'appId=cid-v1:d66f4c7a-ca6e-47bf-b188-9c52589e0f0d'
  },
  body: ReadableStream { locked: false, state: 'readable', supportsBYOB: true },
  bodyUsed: false,
  ok: false,
  redirected: false,
  type: 'basic',
  url: 'https://gosqasbe.azurewebsites.net/api/provenance/totallygarbage'
}
```
We want to instead return a 4XX error response, preferably 404. This is to inform users that they are attempting to reach an API endpoint that doesn't exist.

agilpatr:
We were getting status code 500 when fetching the url "https://gosqasbe.azurewebsites.net/api/provenance/totallygarbage". This is because the path is technically correct (since in provenance/{deviceKey} to reach the endpoint deviceKey just needs to have a value) and when decodeKey threw a custom error without a status code it defaulted to 500.

I've updated the code to throw status 400 (bad request, since the endpoint is correct it's just the key that's invalid) when decodeKey gets an invalid key. I've done the same for postProvenance as well (since it has the same {deviceKey} issue). **Note**: If we do decide we want this to throw 404 instead of 400 just message me either here or on Discord and I can update it.

**Note: Leaving this in draft until #1135 is merged (since the getProvenance and postProvenance specs from that PR will have to be updated to match the new 400 error code).**

Below is what the page now looks like when the key is invalid.
Screenshot 2026-07-13 at 8 24 57 PM


paulErdos:
@agilpatr: #1135 merged. Take a look at the failing check, here: https://github.com/gosqasorg/asset-provenance-tracking/actions/runs/32301978961/job/96226375489?pr=1135

It's in offline mode. I'm not sure if this code will be changed following the new layout. But others will run into this error (blocking their test runs); go ahead and put in a fix under this ticket.

agilpatr:
Hey @paulErdos, that failing check should be fixed now (I gave it more time to run). I re-ran the tests a couple times and they passed every time on my end.

packages/backend/src/functions/httpTrigger.ts
```diff
@@ -98,7 +98,7 @@ export function decodeKey(key: string): Uint8Array<ArrayBuffer> {
         case 32:
             return theKey as Uint8Array<ArrayBuffer>
         default:
-            throw new Error(`Invalid Key Length ${theKey.length}`);
```
paulErdos, line 101: I fear this may turn out to be a breaking change. Several things rely on decodeKey, e.g., in convertLegacy and calculatedeviceid, 
```
    key = typeof key === 'string' ? decodeKey(key) : key;
```

Calculate device id will either throw an uncaught error when sha256 is called on an array, or happily hash the empty array and return the hash, causing in e.g., upload, a nonexistent blob storage location to be queried, throwing an error.

I would do one of two things:
A) Make sure that every call to decodeKey is able to tolerate this new change, or
B) Stick with the original error-throwing version, and add a try/catch to every call to decodeKey. 

The first sounds like less work, although it requires a more detailed look into the mechanics, whereas the second seems like more work, and is somewhat uglier, but is more certain. Both are good.

paulErdos (changes requested):

agilpatr:
Hey @paulErdos, I've gone with the second option and updated decodeKey to throw errors again. To still allow us to display 400 when the key is invalid, I had the endpoints that have deviceKey in their url check if the length is 22 (which is the length of device keys), and return 400 if it's not. I tested every endpoint locally and am getting the correct error handling.

packages/backend/src/functions/httpTrigger.ts
```diff
@@ -461,14 +468,19 @@ async function notifySubscribersHandler(request: HttpRequest, context: Invocatio
...
+        return { status: 400, body: "HTTP Error 400: Invalid Key Length." };
+    }
     const body = await convertLegacyProvenance(containerClient, deviceKey);
     return { jsonBody: body ?? { "already-converted": true} };
 }
 
 export async function getAttachment(request: HttpRequest, context: InvocationContext): Promise<HttpResponseInit> {
     const decryptedBlob = await getDecryptedBlob(request, context);
-    if (!decryptedBlob) { return { status: 404 } }
+    if (!decryptedBlob) { return { status: 400, body: "HTTP Error 400: Invalid Key/AttachmentID Length." } }
```
paulErdos: Hey @agilpatr, plz halp me understand the reasoning here and in getAttachmentName. 

Getting an undefined response from getDecryptedBlob as it is written seems to indicate 404, not 400. 

It seems clearer to me to check request.params.deviceKey: there's no lack of access to it. 

As a separate matter, I'm not seeing a check for attachment id length. I'm not sure how we can say the attachment id length is incorrect based on the existing checks.

packages/backend/src/functions/httpTrigger.ts
paulErdos: deviceID is used only in this log message. DeviceKey allows us to get it. We can drop the calculation of deviceID, and also drop it from the log.

packages/backend/src/functions/httpTrigger.ts
paulErdos: I've spent, probably too long thinking about this. getDecryptedBlob returns undefined in several ways, all of which indicate 404, whether that is due to the user passing bogus information or not (whether these are truly distinct concepts is debatable, they can be considered identical).

As a separate matter, we perform no check for attachment id length. So we can't say anything about this.

I would go back to the 404 here, and on line 499.

paulErdos (changes requested):

PR #1160, opened 2026-07-24
Self-host Font Awesome instead of fetching it from Cloudflare's CDN

### LINKED ISSUE #975: Bring Externally-referenced CSS in-house
We currently fetch a font style sheet from Cloudflare each time we load the history page. Ian's flagged this as a potential security vulnerability, so we'd like to save the file to the repository to avoid issues with incoming code being altered by an attacker. 

https://github.com/gosqasorg/asset-provenance-tracking/blob/625b7de12ff3a4094e4b7b5211a9a080c84fe7f0/packages/frontend/pages/history/%5BdeviceKey%5D.vue#L42

This leaves us vulnerable to CSS injections.

We want to use our own style sheet (possibly just use our own copy of [Cloudflare's style sheet](https://cdnjs.cloudflare.com/ajax/libs/font-awesome/4.7.0/css/font-awesome.min.css) as it's a small file)

CAI-yw:
### Summary
The history page fetched Font Awesome's CSS live from a third-party CDN on every load, with no integrity check. If that response were ever tampered with, attacker-controlled CSS would be injected directly into the page for users.
Vendoring the file removes the runtime dependency on an external origin entirely.

- Downloaded Font Awesome 4.7.0's `font-awesome.min.css` and its 5 webfont files (`.woff2`, `.woff`, `.ttf`, `.eot`, `.svg`) from Cloudflare and committed them under `packages/frontend/public/font-awesome/{css,fonts}/`, preserving the original `css/` + `fonts/` relative layout so the CSS's existing `../fonts/...` `@font-face` references resolve unchanged.
- Updated the `<link>` tag in `pages/history/[deviceKey].vue`(Line42) to point at the local copy (`/font-awesome/css/font-awesome.min.css`) instead of `cdnjs.cloudflare.com`.
- Consideration of vendoring all 5 font formats not just `.woff2`:  Only `.woff2` is actually requested by any modern browser (verified via Network tab), while `.eot`/`.woff`/`.ttf`/`.svg` are legacy fallbacks for very old browsers (IE, pre-2015 Safari/Firefox/Chrome, old Android/iOS webviews) that this Nuxt 3 SPA doesn't realistically need to support. We pulled all 5 anyway for exact parity with what Cloudflare served, so we aren't silently narrowing browser support versus the status quo. But let me know if we decide to only include `.woff2` format. 

### Verification screenshots
`npm run dev` in `packages/frontend`, open the DevTools **Network** tab: 
- Filter by `font-awesome`, hard-refresh a history page, e.g. `/history/<deviceKey>`. Confirm the CSS now loads from `localhost:3000/font-awesome/css/...` with `200`, and **no** request to `cdnjs.cloudflare.com` appears.
- Switch to a mobile viewport and reload, filter Network by `webfont` and confirm `fontawesome-webfont.woff2` loads locally with `200`.

Screenshot 2026-07-23 at 5 06 48 PM
Screenshot 2026-07-23 at 5 17 47 PM


CAI-yw:
It seems that the`[frontend-deploy]` job has succeeded.

paulErdos:
Excellent work and communication as usual, @CAI-yw.

paulErdos (approved):

PR #1171, opened 2026-07-30
Moved "send to all children" Feature from createGroup to addEntry

### LINKED ISSUE #1159: Bring the "send to all children" opertaion into /addEntry
Recently we've decided from a new-user-confusion-minimzation standpoint to both rename "annotate" and to move it from being an operation that can happen during or post group creation to being available strictly post-creation. 

We'd like to bring the API into correspondence with the frontend, and to do that, we'll expand on addEntry to add a "send to all children" boolean field for records that are group parents. If the user hands us the key to a record that is not a group parent, we'll return a 400. 

The twist is that there is some historical momentum in bringing the API up to speed with the prior state of the frontend:

In this ticket, we brought the API into correspondence with the frontend (this was before the change referred to in the first paragraph of this ticket)
Ticket 923: https://github.com/gosqasorg/asset-provenance-tracking/issues/923
PR for ticket 923: https://github.com/gosqasorg/asset-provenance-tracking/pull/1081

Then next, we did a refactor, to use the preexisting notifychildren instead of the custom alternate added in 923:
Ticket 1130: https://github.com/gosqasorg/asset-provenance-tracking/issues/1130
PR for 1130: https://github.com/gosqasorg/asset-provenance-tracking/pull/1137

* Create ticket, or add to above: there are a lot of boolean flags for annotate, e.g. in the signature of notify children, additional instances inside notify children, and in the function signature of createGroup. These should all be removed, for two reasons: the first, that "send to all children" (formerly, "annotate") is no longer a creation operation, and separately, whether annotation is triggered (ie, whether notifyChildren is called) can be predicated on the existence of a datum in the received json handed to addEntry. 
* We'll want to replace the tag value "annotate" and "notify_all" with "sent_to_all_children"

agilpatr:
Moved the "send to all children" feature from createGroup to addEntry (only allowing "send to all children" to be called on records that already exist). Removed the boolean parameters and instead tracked whether or not we're sending to the children by a boolean in the json body called "send_to_all_children" (see the example on the docs page for reference). Changed the "annotate" feature name to "send_to_all_children" in the frontend and backend.

A note for clarity:
- the boolean in the json is called "send_to_all_children"
- and the tag added to record entries is "sent_to_all_children"

Updated api spec and tests to match the new changes, screenshots below show the tests passing locally.
Screenshot 2026-07-29 at 4 49 41 PM
Screenshot 2026-07-29 at 5 07 55 PM
Screenshot 2026-07-29 at 5 08 49 PM


paulErdos (approved):
Excellent work, @agilpatr!

PR #1175, opened 2026-08-06
Fixed Heatmap Data/UI Issues

### LINKED ISSUE #1157: Solve map data + UI issues
Note: This can be worked on at any time, but should be merged only after FabLab26 ends on 31 July
 
* Restore radio buttons and get them working for all time bin granularities the data supports.  
* Pins are showing no data. To clarify, clicking on the pin has the map zoom in, and the geographic tile underlying is not being filled in. It is possible we do not have map data, and for most places, probably nobody minds. For the West Bank / Gaza / Israel area, this can be expected to matter a lot, as hospitals in the area use our services. One odd heisen-bug: clicking on the pin in the West Bank / Gaza / Israel area, then hovering over the pin will present different hover text on a basis that has yet to be ascertained: "West Bank: 186" and "Gaza: 186" have both been seen. I believe those 186 requests are coming from a hospital in Gaza that uses our services (the data will say), and being able to show the region on the map, or at very least to provide stably correct hover text, should be expected to matter a lot to our users in the region. 
* Data not showing in bottom row-tray (previously commented out)

agilpatr:
Made requested changes to the heatmap:
- Restored the radio buttons and fixed the bug where the map would be completely grey when the time was set lower than 3 months. I did this by creating static counts for each time period (instead of trying to filter recent-requests.json, which only contains 20 records)
- Got pins to show data from recent-requests.json file
- Restored the bottom-tray and got data to appear. Bottom tray gets its data from the recent-requests.json file, not live data, which is why it only displays a few requests per region (does correctly hide requests if they’re not in the selected time range).

I also updated the country counts to exist for all time periods, but instead of doing the same for region/city I modified the code to display the number of recentRequests when hovering (instead of the static counts). This way the region/city colors, hover, and bottom tray all match.
I didn’t update the requests in recent-requests.json since I didn’t want to leak actual records, and wasn’t sure exactly how we wanted to go about filtering the data.

### DIFF
--- a/packages/frontend/data_viz/files/requests-by-country@5.json
+++ b/packages/frontend/data_viz/files/requests-by-country@5.json
@@ -1,26 +1,56 @@
 [
   {
     "country": "USA",
-    "requests": 4654
+    "one_year": 18661,
+    "six_months": 18661,
+    "three_months": 18661,
+    "one_month": 4789,
+    "one_week": 1718,
+    "one_day": 0
   },
   {
     "country": "CAN",
-    "requests": 1430
+    "one_year": 770,
+    "six_months": 770,
+    "three_months": 770,
+    "one_month": 90,
+    "one_week": 0,
+    "one_day": 0
   },
   {
     "country": "FRA",
-    "requests": 179
+    "one_year": 91,
+    "six_months": 91,
+    "three_months": 91,
+    "one_month": 8,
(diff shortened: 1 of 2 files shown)


paulErdos (approved):
@agilpatr Excellent work. Great thinking on using .includes.

PR #1176, opened 2026-08-06
Fixed Zod Validation

### LINKED ISSUE #1164: [BUG] Zod validation of record JSON is broken
Zod schema validation with z.object({...}).parse(json) is synchronous. Our validateJson function is written as async, but is not awaited when called: the function returns a promise object, and promises are truthy. The result of the function call is never resolved, and execution passes regardless of the function call's return value: if validateJson detects an error, nothing happens: executions just rolls on by. 

That is the first bug.

The second bug appears when the first bug is resolved, either by awaiting, or removing async in validateJson's signature. 

The bug itself is a unit test failure in test/UnitTests/httpTrigger.endpoints.spec.ts on line 156. The issue may be that postProv is being handed malformed input, and that the schema validator is working as intended. 

Some other cleanup:
* Zod validation is going to be used elsewhere in the code. Let's rename the function to something non-generic, e.g., 'validateRecordJson'.
* Let's drop the calls in convertLegacyProv and getProv.

agilpatr:
Fixed zod validation based on issue comments:
- Removed “await” from the validate function
- Fixed failing test by using a correctly formatted record
- Renamed the validation function to validateRecordJSON to be less generic
- Removed zod validation in covertLegacyProvenance and getProvenance

Also made some changes to the code (since createGroup/addEntry were failing due to validation):
- Removed “children_key” as a required field, since postProvenance is validating both new records (which need children_key) and record entries (which does not). Now only description is required for a valid record
- Removed tests that required empty description (Note: New entries used to allow empty descriptions, but with this change they no longer do. It might be better to validate new records and new entries separately, as they have different requirements)


paulErdos:
Hey @agilpatr! Let's keep the test. We do still as a feature allow users to create records with empty descriptions and send them to all children. Someone may want to send to all children just a tag, for example, e.g., "Received" or "Defective".

If this leaves a description-sized hole in our postProvenance validation, then that may be okay. That being said, within postProvenance, it may be possible to distinguish between new provenance creations, and record history element additions, especially in the context of sending to all children, by inspecting the json itself. A helper function that distinguishes the two cases and applies the corresponding schema would be valuable.

agilpatr:
Hey @paulErdos, I made the changes you requested! The code now determines whether we're looking at a record/group or a record entry based on if the json contains deviceName (a field that only records/groups should have), and validates accordingly. I also added new tests for entry validation.

packages/backend/src/functions/httpTrigger.ts
```diff
@@ -410,7 +407,12 @@ export async function postProvenance(request: HttpRequest, context: InvocationCo
     const provenanceRecord = formData.get("provenanceRecord");
     if (typeof provenanceRecord !== 'string') { return { status: 404 }; }
     const record = JSON5.parse(provenanceRecord);
-    if (!validateJSON(record)) { return { status: 404 }; }
+
+    if ("deviceName" in record) {
+        if (!validateRecordJSON(record)) { return { status: 400 }; }
```
paulErdos, line 412: In general, a 400 error is a message to the user that they did something wrong. 

If we don't tell them what they did wrong, then it's like if I were to start a review on this line saying nothing more than "Something here is amiss". 

Typically for an edit so small I'd just make the change, and in not doing so I'm creating more work for myself. Regardless I am a sucker for education, so FYI also plzfix

Image

packages/backend/src/functions/httpTrigger.ts
```diff
@@ -706,6 +708,24 @@ export async function validateJSON(json: any) {
     }
 }
 
+export function validateEntryJSON(json: any) {
```
paulErdos, line 711: This has been so needed for so long. 

No edits needed here. Just a comment.

packages/backend/test/UnitTests/httpTrigger.endpoints.spec.ts
```diff
@@ -177,46 +187,75 @@ describe('httpTrigger endpoints (shallow mocks)', () => {
...
+    let valid = await httpTrigger.validateRecordJSON(invalidName);
+    expect(valid).toBe(false);
+
+    // Description is missing, which should cause validateRecordJSON to flag this record as invalid
+    const invalidDescription = {"blobType":"deviceInitializer","deviceName":"No Description",
+      "tags":["record"],"children_key":"","hasParent":false,"isPublicKey":false};
+    valid = await httpTrigger.validateRecordJSON(invalidDescription);
+    expect(valid).toBe(false);
+
+    // hasParent is of the wrong type, which should cause validateRecordJSON to flag this group as invalid
```
paulErdos, line 232: Nicely done. This is a very expectable user error. 

No edits needed. Just a comment.

paulErdos (changes requested):
Just the one change related to the messageless 400s. All other changes reviewed.

agilpatr:
Hey @paulErdos, good catch, I appreciate the learning opportunity! I've updated the code to return an error message "Format of provided JSON is invalid" along with status code 400.

paulErdos (approved):

PR #1177, opened 2026-08-08
Forked Email Send

### LINKED ISSUE #1154: Fork email send to reduce record creation lag
Creating records in provenance histories, ie, adding record history elements, when users have signed up for notifications lags.

Presently, we, in short (starting from a history page):

0. User clicks "create record"
1. Fire off postProvenance
2. Transition to pending state page
3. Fire off email send
4. Wait for email send to complete, ie, await is used in the function call
5. Fire off getProvenance
6. Transition back to history page

When emails send successfully, we display nothing in the snackbar. When emails fail to send, we show a message in the snackbar. 

What we'd like to do is not await. The best strategy may be to create an async function that sends the call and manages the snackbar, and when it is called, simply don't await it. Then execution will fork.

agilpatr:
Created a new endpoint notifySubscribers that sends emails to all subscribers of that key. This is so the frontend can separately call this new endpoint (without await) and upon it's return display either success or failure. If emails are successfully sent then success is shown, if no emails are subscribed then nothing is displayed.

Also removed the notifySubscribers call in postProvenance and instead called the function in recall, send to all children, and addEntry.


paulErdos (approved):
Approved! Well done. Very good going above + beyond adding calls to notifysubscribers where needed. I really like the "successfully emailed subscribers" snackbar message: seeing it pop up after a delay was great. 

Note here in createrecord line 260:

``` 
let errorMessage;
if (error instanceof Error) {
    errorMessage = error.message;
} else {
    errorMessage = error;
}

this.$snackbar.add({
    type: 'error',
    text: `Error sending email: ${errorMessage}`
});
```

this could be replaced with one ternary expression, even in the text field of snackbar.add, e.g.:
```
text: `Foo ${ error instanceof Error ? error.message : error}`
```

This saves a few lines and a variable. I'm going to approve this now: add a commit containing that change, if you like (up to you), test it out (deploy to dev) to make sure it works, and merge it in. 

Ternary operator docs here:
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Conditional_operator

PR #1193, opened 2026-08-17
added validation for api endpoints

### LINKED ISSUE #1163: Add Zod JSON Validation to all APIv2 Endpoint Handlers
When using the frontend, the user is not at risk of sending malformed json to the backend.

When using the API, that risk exists. Presently, we are not set up to let the user know, by sending http status code 400, that their request was malformed. 

For an example of how to use Zod, take a open httpTrigger.ts and look at RecordCreationOrderSchema's definition and use around lines 1573-1593 as well as the zod-specific error handling in the catch block of createRecordHandler. 

Links:
* https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Status/400

deneille:
- Piggbacked off `RecordCreationOrderSchema` pattern to add Zod schema validation to every API v2 endpoint handler except postProvenance (See PR #1176 for that implementation by @agilpatr). Now each endpoint should validate its inputs and return 400 error.

- Verified all tests against unit tests and integration tests passed, and updated tests where needed. Also added an integration test that checks for malformed or invalid data .N.B.: `create.test.ts` valid `should fail to create a group record with an invalid device key` integration test should be 400 but held off because of `postProvenence` schema validation was implemented in PR #1176.

- Added api endpoints in `docs` swagger specs that did not have them.


paulErdos:
Hey @deneille! Reviewing this, I'm not seeing a schema for postProvenance. No need for a deviceKeySchema. Basically for functions that have a signature that looks like this:

```
export async function getDecryptedBlob(request: HttpRequest, context: InvocationContext): Promise<DecryptedBlob | undefined> 
```

We want to have a schema to validate the contents of the request parameter.

deneille:
Hey @paulErdos made the changes you asked for. I just thought PR #1176 had some overlap with `postProvenance`  validation, regardless I added the schema. Let me know if I need to fix something else.

paulErdos:
@deneille Right on. Thanks for that. This one's a good amount to review so still work in progress from my end.

paulErdos:
Merging main: checks should be free of edgesOut error now.

packages/backend/src/functions/httpTrigger.ts
```diff
@@ -369,120 +372,191 @@ async function countExistingAttachments(containerClient: ContainerClient, device
...
+            status: 500,
+            jsonBody: { message: "Internal Server Error" },
+        }
     }
-    records.sort((a, b) => b.timestamp - a.timestamp)
-    return { jsonBody: records };
 }
 
-export async function postProvenance(request: HttpRequest, context: InvocationContext): Promise<HttpResponseInit> {
+const PostProvenanceSchema = z.object({
```
paulErdos, line 416: Go ahead and move function-specific schemata inside their functions.

packages/backend/src/functions/httpTrigger.ts
```diff
@@ -783,30 +865,38 @@ async function addRecordWithTags(baseUrl, deviceKey, tags, description) {
     });
 }
 
+const RecallTagsDescriptionSchema = z.object({
```
paulErdos: Nice.

packages/backend/src/functions/httpTrigger.ts
```diff
@@ -816,46 +906,62 @@ export async function recall(request: HttpRequest, context: InvocationContext):
...
+                status: 400,
+                jsonBody: { message: "Error: Check argument format." },
+            }
         }
+        return { status: 500 };
     }
 }
 
+const EmailVerificationSchema = z.object({
+    email: z.email(),
```
paulErdos, line 962: Nice that they have this. Parsing an email with eg regex is to my understanding not possible.

paulErdos:
Very thorough. This is great work. Still finishing review.

packages/backend/src/functions/httpTrigger.ts
```diff
@@ -369,120 +372,191 @@ async function countExistingAttachments(containerClient: ContainerClient, device
...
+        records.sort((a, b) => b.timestamp - a.timestamp)
+        return { jsonBody: records };
+    } catch (error) {
+        if (error instanceof z.ZodError) {
+            return {
+                status: 400,
+                jsonBody: { message: "Error: Check argument format of device key." },
+            }
+        }
+        return {
```
paulErdos: Make this an else

packages/backend/src/functions/httpTrigger.ts
```diff
@@ -315,10 +315,13 @@ async function convertLegacyProvenance(containerClient: ContainerClient, key: Ui
     return records;
 }
 
+const AttachmentIDSchema = z.string().regex(/^[0-9a-f]{64}$/);
+
 export async function getDecryptedBlob(request: HttpRequest, context: InvocationContext): Promise<DecryptedBlob | undefined> {
     const deviceKey = decodeKey(request.params.deviceKey);
     const deviceID = await calculateDeviceID(deviceKey);
     const attachmentID = request.params.attachmentID;
+    AttachmentIDSchema.parse(attachmentID);
```
paulErdos, line 324: Go ahead and handle this at the beginning of the try blocks of getAttachment and getAttachmentName.

packages/backend/src/functions/httpTrigger.ts
```diff
@@ -369,120 +372,191 @@ async function countExistingAttachments(containerClient: ContainerClient, device
...
+            // if (!(await validateJSON(json))) { return { status: 400 }; }
+            // validateJSON is broken
+            const parsed_json = JSON.parse(json);
+            const provRecord = parsed_json as ProvenanceRecord;
+            records.push({ ...provRecord, deviceID, timestamp });
+        }
+        records.sort((a, b) => b.timestamp - a.timestamp)
+        return { jsonBody: records };
+    } catch (error) {
+        if (error instanceof z.ZodError) {
```
paulErdos, line 403: This function is not doing any parsing, so we can drop this conditional from the catch block.

packages/backend/src/functions/httpTrigger.ts
```diff
@@ -369,120 +372,191 @@ async function countExistingAttachments(containerClient: ContainerClient, device
...
-                error: 'Failed to send email'
+        return { jsonBody: body ?? { converted: true}};
+    } catch(error){
+        if (error instanceof z.ZodError) {
+            return {
+                status: 400,
+                jsonBody: { message: "Error: Check argument format." },
             }
         }
+        return {
```
paulErdos, line 490: Let's put this in an else to correspond to the above if.

packages/backend/src/functions/httpTrigger.ts
```diff
@@ -369,120 +372,191 @@ async function countExistingAttachments(containerClient: ContainerClient, device
...
-    const deviceKey = decodeKey(request.params.deviceKey);
-    const body = await convertLegacyProvenance(containerClient, deviceKey);
-    return { jsonBody: body ?? { "already-converted": true} };
+    try{
+        const deviceKey = decodeKey(request.params.deviceKey);
+        const body = await convertLegacyProvenance(containerClient, deviceKey);
+        return { jsonBody: body ?? { "already-converted": true} };
+    } catch (error) {
+        return {
+            status: 500,
```
paulErdos, line 504: Nice work putting this in a try/catch. Not sure how this went unaddressed for so long: thanks for adding it. 

Taking a look at the error that is thrown, decodeKey throws if it gets a malformed deviceKey, which ultimately comes from the user. Let's return 400, and use the same error message from the thrown error, which is here:

https://github.com/gosqasorg/asset-provenance-tracking/blob/13d62830bbe92c85cee089bf0f29635ff018c63c/packages/backend/src/functions/httpTrigger.ts#L100C17-L101C68

paulErdos (changes requested):

packages/backend/src/functions/httpTrigger.ts
```diff
@@ -369,120 +372,191 @@ async function countExistingAttachments(containerClient: ContainerClient, device
...
-    const { data, contentType, filename } = decryptedBlob;
-    const headers = new Headers();
-    headers.append("Access-Control-Allow-Headers", "Attachment-Name");
-    if (contentType) { headers.append("Content-Type", contentType); }
-    if (filename) {
-        headers.append("Content-Disposition", `attachment; filename="${filename}"`);
-        headers.append("Attachment-Name", filename);
-    }
+    try{
+        const decryptedBlob = await getDecryptedBlob(request, context);
```
paulErdos, line 512: Okay, so I see you're allowing a catch for the error thrown by getDecryptedBlob here. Probably best to leave this as is, as endpoints should be written with a try/catch by default.

I'm seeing that although getDecryptedBlob takes a request and a context, it's not registered as an endpoint. 

It is true that we'd be doing validation in two places as opposed to one. So your work does feature an optimal solution. However, the principle to follow is to validate at trust boundaries, i.e., the point of entry. 

As a separate matter: go ahead and also inspect the zod error text itself. If it is useful, go ahead and hand it back in the catch block's if instance of zod error section. The causes of 500 errors are something we want to be opaque to the user about (although verbose in logs). 400 errors though, informing the user is good. 

Note: it's just a one line change, nbd.

packages/backend/src/functions/httpTrigger.ts
```diff
@@ -369,120 +372,191 @@ async function countExistingAttachments(containerClient: ContainerClient, device
...
+            status: 500,
+        }  
+    }
 };
 
 export async function getAttachmentName(request: HttpRequest, context: InvocationContext): Promise<HttpResponseInit> {
-    const decryptedBlob = await getDecryptedBlob(request, context);
-    if (!decryptedBlob) { return { status: 404 } }
+    try{
+        const decryptedBlob = await getDecryptedBlob(request, context);
```
paulErdos, line 542: Same request as on line 512. It's just a one-line change.

packages/backend/src/functions/httpTrigger.ts
```diff
@@ -783,30 +865,38 @@ async function addRecordWithTags(baseUrl, deviceKey, tags, description) {
...
-    const formData = await request.formData();
-    const recordStr = formData.get("provenanceRecord"); 
-    const record = JSON5.parse(formData.get("provenanceRecord") as string) || { tags: []};
-
-    record.tags ??= [];
-    if (!record.tags.includes("recall")) record.tags.push("recall");
-    const tags = record.tags
     
-    const description = record.description || "";
+    try{
```
paulErdos, line 877: Let's do the validation at the top of the try block. For recall, the parameter is just device key. 

Validate key simply asks whether it's been passed a length 22 string made of only lowercase letters, uppercase letters, and numbers. This would save the work of altering the catch block logic. It would also not break the pattern by creating a special case. 

And it looks like you do that, line 1008. Use that strategy here.

packages/backend/src/functions/httpTrigger.ts
```diff
@@ -868,10 +974,9 @@ export async function postEmail(request: HttpRequest, context: InvocationContext
         await tableClient.createTable();  // Create if not exist, no error if it does
 
         const formData = await request.formData();
-        let email; if (typeof (email = formData.get('email')) !== 'string') {
-            throw new Error('postEmail: Unexpected non-string value received')
-            return { status: 404 };
-        }
+        const emailStr = formData.get('email');
+        let theRequest = EmailVerificationSchema.parse({ email: emailStr } );
+        let email = theRequest['email'];
```
paulErdos, line 979: Very nice.

packages/backend/src/functions/httpTrigger.ts
```diff
@@ -888,26 +993,31 @@ export async function postEmail(request: HttpRequest, context: InvocationContext
...
         }
     } catch(error) {
-        console.error('postEmail: Failed to add feedback volunteer contact info', error.message)
+        if (error instanceof z.ZodError) {
+            return {
+                status: 400,
+                jsonBody: { message: "Error: Check argument format." },
+            }
+        }
+        context.error('postEmail: Failed to add feedback volunteer contact info', error.message)
```
paulErdos, line 1002: Very nice correcting the console.error to context.error!

packages/backend/src/functions/httpTrigger.ts
```diff
@@ -1228,29 +1362,36 @@ export async function postResendCode(request: HttpRequest, context: InvocationCo
     }
 }
 
+const NotificationUnsubscribeSchema = z.object({
+    id: z.string().min(43).max(44).regex(/^[1-9A-HJ-NP-Za-km-z]+$/),
```
paulErdos, line 1366: Just want to check the regex here: tokens are a-zA-Z0-9, except "I", "M", "O", and "l", is how I'm reading this. Not wrong, to be clear. But seems unusual. I have not looked at token generation: do they indeed follow this exclusion pattern? 

May be worth a comment. 

Since this is only used in one place, go ahead and move it inside deleteNotificationEmail, above the try block.

packages/backend/src/functions/httpTrigger.ts
```diff
@@ -1645,66 +1786,91 @@ export async function createRecordHandler(request: HttpRequest, context: Invocat
...
-
-    // Check the first record entry in the provenance to see if the key is a group or not
-    const provenance = await getProvenance(request, context);
-    const creationRecord = provenance.jsonBody[provenance.jsonBody.length - 1];
-    if (!creationRecord) {
-        return {
-            status: 400,
-            jsonBody: { error: "Provenance needs to exist before adding entries." }
+    const backendUrl = process.env
+    ['backend_url'];
```
paulErdos, line 1806: It looks like this is meant to go on the line above.

packages/backend/src/functions/httpTrigger.ts
```diff
@@ -1645,66 +1786,91 @@ export async function createRecordHandler(request: HttpRequest, context: Invocat
...
+    try {
+        const requestClone = request.clone();
+        const deviceKey = requestClone.params.deviceKey;
+        let formData = await requestClone.formData();
+        const attachmentValues = formData.values();
+        const record = JSON.parse(formData.get("provenanceRecord") as string);
+        AddEntrySchema.parse(record);
```
paulErdos, line 1813: What is being parsed here is a record. It may be valuable to keep the schema, renamed, and applied elsewhere. Possibly not with all-optional, unless there is a reason for doing so. If there is, perhaps it is two different use cases of record, e.g., an entry, or the entire provenance. 

But all that's needed for addEntry is to check that the deviceKey is passed correctly.

paulErdos (changes requested):

packages/backend/test/IntegrationTests/Live/v1/create.test.ts
```diff
@@ -1021,7 +1021,7 @@ describe("Group Creation Tests", () => {
 		// Expect the response to indicate failure (e.g., 500 Internal Server Error)
 		console.log("Response Status for Invalid Key Test: " + groupResponse.status);
 		expect(groupResponse.ok).toBe(false);
-		expect(groupResponse.status).toBe(500);
+		// expect(groupResponse.status).toBe(400); //Should be uncommented when merged with Alison's code
```
paulErdos, line 1024: Note: I believe Alison's code has been merged. If not, let me know which PR you need me to prioritize.

packages/backend/test/IntegrationTests/Live/v2/create.test.ts
```diff
@@ -447,6 +447,28 @@ describe("Group Creation Tests", () => {
     }
     }, 60000);
 
+    it("should reject an invalid group creation payload with a 400", async() => {
+        const invalidPayload = {
+            deviceName: "invalid_group_test_payload",
+            description: "this payload has tags as string",
+            tags: "[tag, string, invalid]",
```
paulErdos, line 454: Great test case.

paulErdos (changes requested):

paulErdos:
@deneille 

Golly! This is a big one. Don't worry about the volume of requested changes, go at your own pace. At the end I'm happy to help resolve merge conflicts / etc.

packages/backend/src/functions/httpTrigger.ts
deneille: It is because base58 omits those characters. When I checked this, the email id being generated uses base58encode in the emailNotificationUtil file. I can just take off that distinct and only check for alphanumeric strings. Here it talks about base58 encoding.

Let me know what you think.

PR #1202, opened 2026-08-24
needs to be tested but fixed QR code url on dev site

### LINKED ISSUE #1196: Bug: QR code not picking up non-local urls correctly
Pic related. The QR code points to the url displayed. I've taken a look into where the QR picks up the url and it's not clear what is going wrong. 

Image

deneille:
This bug seems to be related to dev site exclusively. When the frontend dev is deployed, `FRONTEND_URL=/` . Hence, why `/` is seen as the url of origin for the frontend. 

Still need to test on dev to confirm that this fix works.

### DIFF
--- a/packages/frontend/components/QRCode.vue
+++ b/packages/frontend/components/QRCode.vue
@@ -23,6 +23,9 @@ along with this program.  If not, see <https://www.gnu.org/licenses/>. -->
 import { useNuxtApp } from '#app';
 import type QRCodeStyling from 'qr-code-styling';
 
+// QR Code Image
+import QRCodeImage from './gosqas_main_logo_white_border_64BitEncodedPng.txt?raw' with { type: 'text' };
+
 // Styling for custom text
 const CUSTOM_TEXT_MAX_LENGTH = 100;
 const CUSTOM_TEXT_FONT = 'bold 20px Arial, sans-serif';
@@ -62,7 +65,12 @@ export default {
         width: 322,
         height: 361,
         type: 'canvas',
+        image: "data:image/png;base64," + QRCodeImage,
         data: this.url,
+        qrOptions: {
+          typeNumber: 10,
+          errorCorrectionLevel: 'H'
+        },
         imageOptions: {
           hideBackgroundDots: false,
           imageSize: 1,
--- a/packages/frontend/components/QRCode.vue
+++ b/packages/frontend/components/QRCode.vue
@@ -21,9 +21,6 @@ along with this program.  If not, see <https://www.gnu.org/licenses/>. -->
 import { useNuxtApp } from '#app';
 import type QRCodeStyling from 'qr-code-styling';
 
-// QR Code Image
-import QRCodeImage from './gosqas_main_logo_white_border_64BitEncodedPng.txt?raw' with { type: 'text' };
-
 // Styling for custom text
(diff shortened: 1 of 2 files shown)


paulErdos:
@deneille Nice solution! Very creative not relying on the env var.

I'm seeing on dev, there is no env var for the frontend url. That may be because this is the way to do it. 

I'll go ahead and deploy to test.

paulErdos:
Curious that it's seen as dev exclusively: there are no env vars for prod.

paulErdos:
Huh. Strange. It's still not working.

paulErdos:
More odd phenomena: 
Screenshot 2026-08-23 at 21 01 23

packages/frontend/components/QRCode.vue
```diff
@@ -37,7 +34,14 @@ const URL_TEXT_COLOR = '#1E2019';
...
 const QR_SCALE_WITH_CUSTOM_TEXT = 0.9; // Scale down QR code when custom text is shown
 
-
+function normalizeUrl(url: string): string { 
+  /* when dev is deployed the frontend url is `/` which is why there is a bug for qrCode on dev */ 
+  if(/^https?:\/\//i.test(url)){
+    return url;
+  }
+  const cleanPath = url.replace(/^\//, "");
+  return `$window.location.origin/${cleanPath}`;
```
paulErdos, line 48: $window is not a thing in either vue or nuxt.

packages/frontend/components/QRCode.vue
```diff
@@ -92,7 +95,7 @@ export default {
   watch: {
     url(newValue: string | undefined) {
       if (newValue) {
-        this.options.data = newValue;
+        this.options.data = normalizeUrl(newValue);
```
paulErdos, line 109: We don't want this to be set in watch. It does not change.

packages/frontend/components/QRCode.vue
paulErdos, line 54: url should not be a prop.

paulErdos: The url that is defined as a member of options, named 'data' (see line 113), should be used here.

paulErdos, line 261: Hardcoded filename. Should be, e.g., ```GDT-<device key>```

paulErdos (changes requested):

paulErdos:
Take a look at the commit I added, commit message 'Resolve url issue'. That solves the problem, the url displays correctly. 

There are still revisions needed, but that was the issue.

packages/frontend/components/QRCode.vue
paulErdos: Remember to remove all references to this.url

paulErdos: Update: do not set the url in either watch url or in data.

paulErdos: Also remember to remove the prop being passed in the code that imports this (modals/qrcode) as well as in history+record/devicekey, and drop the qrcode url variable from these as well.

paulErdos:
@deneille Take a look at the example, showing how to download near the bottom, on the npm page for the qr code styling library:
https://www.npmjs.com/package/qr-code-styling

paulErdos:
Solved under ticket #1208

PR #1203, opened 2026-08-24
889 feature request security and ethics feature  skin and face detection

### LINKED ISSUE #889: [FEATURE REQUEST] Security and Ethics feature- skin and face detection
To safeguard our server against illicit use, we should implement a media flesh-detection algorithm and face blurring in our code. We should consider an option for communication with a user, as they may try to photograph a medical device that has a body part in frame (an external fixator attached to a limb). Is there a way to send an alert "please contact us as your uploads are flagging GDT security parameters"? If this is possible, we could build a clearlist of users. However, users may not contact us after alert, and if so they would be blocked from using the system if usage was not explained.

This may be slightly related to [Issue #37](https://github.com/gosqasorg/asset-provenance-tracking/issues/37)

@paulErdos tagging you for assignment at your discretion
@JaraR tagging you for budget visibility, as this will affect our resource use. If possible, we should try to ballpark an annual budget in the idea of "if we have 10,000 photo records/year, what will the resource costs be?"

This issue may require further team discussion for implementation clarity.

paulErdos:


paulErdos:
Note to self: drop the throw error on credentials not being set.

PR #1200, opened 2026-08-19
Remove Outdated Offline Functions

HieuNgoGH:
Removed these functions from azureFuncs.ts and their corresponding tests in azureFuncs.spec.ts
-  onlineTestFetch
- connectivityChecker
- stashRequest (not stashOfflineRequest)
- emptyStash
- periodicChecker
- offlineDetectAndStash

PR #1204, opened 2026-08-24
Removed Final Outdated Offline Variables

agilpatr:
A couple of offline variables still needed to be removed to allow the frontend to work as normal (currently gdt/history pages won't load on dev). I removed those variables/function calls here, as well as removing the syncing stash. Tested the regular frontend features locally and everything should be working as normal.

There are frontend test failures due to not reaching the coverage threshold. The tests themselves do seem to be passing (whereas before they would fail completely).


paulErdos:
Copying this from Hieu's PR, which I foolishly merged without noticing the test failure.

Hey @HieuNgoGH, it looks like neither of us did our due diligence here. We need to remove all of the references, uses, and imports of these functions as well. I should have noticed the errors. Well, we both should have. Human error, no worries, just please take care of it.

@HieuNgoGH, since the other is closed, it may be worthwhile to continue the work in this pr, checking out the branch, merge main, push, and the test fail should reappear. 

@agilpatr seeing a separate test failure in frontend coverage checks. Would you look into it please?

paulErdos:
Looking at the changes, it appears, @agilpatr, that you may have solved the test failures here. Is that the case?

https://github.com/gosqasorg/asset-provenance-tracking/pull/1200

agilpatr:
Hey @paulErdos, yes the errors should be fixed here. Looking at the failing test it reports that all tests pass, however our tests only cover 3.67% of lines while the threshold is 3.9% (which is where the error comes from). I wasn't sure how we wanted to tackle this. This branch does fix the broken pages and the broken tests, but devs will still get the above test failure.

paulErdos:
Greetings @agilpatr. These two docs show how to set the coverage checks to only the files that are changed. The first one shows how to do it as a command, the second one shows how to set *that* in the vitest config. 
1. https://vitest.dev/config/coverage.html#coverage-changed
2. https://vitest.dev/guide/coverage.html

paulErdos (approved):

PR #1205, opened 2026-08-24
Liveness Checks

### LINKED ISSUE #1173: Liveness Checks
Have dev ping prod, have prod ping dev, if either is down, trigger email send.

Emails to send to are in the environment variable "LIVENESS_CHECK_EMAIL_RECIPIENTS", and are a comma-separated string of my email address, and Jara's email address, formatted as, e.g., ```email1@domain.com,email2@domain.com```. 

Do this as a time-trigger function set to run once per hour.

HieuNgoGH:
Created liveness checker function that checks whether dev or prod is down with a timer that checks every hour. If any server is down, an email will be sent to the emails in environment variable "LIVENESS_CHECK_EMAIL_RECIPIENTS".

UPDATE: fixed livenessChecker function and added helper function to send email and log error.

### DIFF
--- a/packages/backend/src/functions/httpTrigger.ts
+++ b/packages/backend/src/functions/httpTrigger.ts
@@ -8,13 +8,14 @@ import JSON5 from 'json5';
 import * as z from "zod";
 import { webcrypto as crypto } from 'node:crypto';
 import { TableClient, AzureNamedKeyCredential } from '@azure/data-tables'
-import { app, HttpRequest, HttpResponseInit, InvocationContext } from "@azure/functions";
+import { app, HttpRequest, HttpResponseInit, InvocationContext, Timer } from "@azure/functions";
 import { BlockBlobClient, ContainerClient, StorageSharedKeyCredential } from "@azure/storage-blob";
 import { VERSION_INFO } from '../version.js';
 import { makeEncodedDeviceKey } from '../utils/keyFuncs.js';
 import { notifySubscribers, retrieveNotifEmails, updateNotifications } from './emailNotificationUtils.js';
 import { ClientSecretCredential } from "@azure/identity";
 import './getStats.js';
+import { sendEmail } from './sendEmail.js'
 
 // To deploy this project from the command line, you need:
 //  * Azure CLI : https://learn.microsoft.com/en-us/cli/azure/
@@ -1708,12 +1709,72 @@ export async function addEntryHandler(request: HttpRequest, context: InvocationC
     }
 }
 
+export async function livenessChecker(livenessTimer: Timer, context: InvocationContext) {
+
+    const staging = 'https://red-stone-00f5d251e.5.azurestaticapps.net/'
+    const production = 'https://blue-stone-05ede120f.5.azurestaticapps.net/'
+    const frontendUrl = process.env['frontend_url']
+    const emails = process.env['LIVENESS_CHECK_EMAIL_RECIPIENTS']
+
+    if (frontendUrl === staging) {
+        const prodResponse = await fetch('production');
+        if (prodResponse.status != 200) {
+            try {
+                for (const email of emails) {
+                    const emailResponse = await sendEmail(
+                        process.env['SENDER_EMAIL'],
+                        email,
+                        'Important: Production Server Down',
+                        'Production server is down!',
+                        'GOSQAS DEVS',
+                        context
+                    )
+                if (emailResponse.status === "Failed") {
+                    throw emailResponse
+                    }
+                }
+            }
+            catch (error) {
+                console.log(error)
+            }
+        }
+    }
+    if (frontendUrl === production) {
+        const stageResponse = await fetch(staging);
+        if (stageResponse.status != 200) {
+            try {
+                for (const email of emails) {
+                    const emailResponse2 = await sendEmail(
+                        process.env['SENDER_EMAIL'],
+                        email,
+                        'Important: Staging Server Down',
+                        'Staging server is down!',
+                        'GOSQAS DEVS',
+                        context
+                    )
+                if (emailResponse2.status === "Failed") {
+                    throw emailResponse2
+                    }
+                }
+            }
+            catch (error) {
+                console.log(error)
+            }
+        }
+    }
+
+}
+
 // Once per day update the total record, record entry, and attachment counts
 app.timer('updateRecordCounts', {
     schedule: `0 0 * * *`,
     handler: setStatisticsTotals
 })
 
+app.timer('livenessChecker', {
+    schedule: '0 0 * * * *', 
+    handler: livenessChecker,
+})
 
 /* ----- API Endpoints Section 2/2: Route Definitions ----- */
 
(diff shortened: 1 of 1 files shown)


packages/backend/src/functions/httpTrigger.ts
```diff
@@ -1707,12 +1709,72 @@ export async function addEntryHandler(request: HttpRequest, context: InvocationC
     }
 }
 
+export async function livenessChecker(livenessTimer: Timer, context: InvocationContext) {
+
+    const staging = 'https://red-stone-00f5d251e.5.azurestaticapps.net/'
```
paulErdos, line 1714: Be sure to include the other two, gosqas.org and dev.gosqas.org. Could then write:
const location = frontendUrl.includes('red') || frontend.includes('dev') ? 'production' : 'staging';

and use that in your first if, and make the second one an else. Better yet, return at the end of the first if, and free the second from the conditional.

packages/backend/src/functions/httpTrigger.ts
```diff
@@ -1707,12 +1709,72 @@ export async function addEntryHandler(request: HttpRequest, context: InvocationC
     }
 }
 
+export async function livenessChecker(livenessTimer: Timer, context: InvocationContext) {
+
+    const staging = 'https://red-stone-00f5d251e.5.azurestaticapps.net/'
+    const production = 'https://blue-stone-05ede120f.5.azurestaticapps.net/'
+    const frontendUrl = process.env['frontend_url']
+    const emails = process.env['LIVENESS_CHECK_EMAIL_RECIPIENTS']
+
```
paulErdos, line 1735: Nothing to do for this one, just have to say, very nicely organized.

packages/backend/src/functions/httpTrigger.ts
```diff
@@ -1707,12 +1709,72 @@ export async function addEntryHandler(request: HttpRequest, context: InvocationC
...
+                        'GOSQAS DEVS',
+                        context
+                    )
+                if (emailResponse.status === "Failed") {
+                    throw emailResponse
+                    }
+                }
+            }
+            catch (error) {
+                console.log(error)
```
paulErdos: In Azure Functions API v4, console is deliberately disabled. Use context for logging, e.g., context.log('foo'), and for errors, use context.error("oh no!")

packages/backend/src/functions/httpTrigger.ts
```diff
@@ -1707,12 +1709,72 @@ export async function addEntryHandler(request: HttpRequest, context: InvocationC
...
+            }
+        }
+    }
+    if (frontendUrl === production) {
+        const stageResponse = await fetch(staging);
```
paulErdos: Move both fetches above both ifs. Then you can compound the nested conditionals into one eg if(frontendurl === production && prodResponse.status != 200)

packages/backend/src/functions/httpTrigger.ts
```diff
@@ -1707,12 +1709,72 @@ export async function addEntryHandler(request: HttpRequest, context: InvocationC
...
+        if (stageResponse.status != 200) {
+            try {
+                for (const email of emails) {
+                    const emailResponse2 = await sendEmail(
```
paulErdos: Since the only difference between the email bodies is "Staging" vs "Production", we can functionalize to simplify the code. 

Write a function that handles the email send, have it take one parameter, and have it parametrize the strings in the arguments to sendEmail. Have it contain the try/catch as well. Pass context to it, and it han handle the error logging. Then you have the logic structure and the error-prone network interaction separate.

paulErdos (changes requested):
Slinging tricks of the trade your way.

packages/backend/src/functions/httpTrigger.ts
```diff
@@ -1707,12 +1709,72 @@ export async function addEntryHandler(request: HttpRequest, context: InvocationC
...
+                        context
+                    )
+                if (emailResponse2.status === "Failed") {
+                    throw emailResponse2
+                    }
+                }
+            }
+            catch (error) {
+                console.log(error)
+            }
```
paulErdos: Go ahead and write two simple unit tests, by e.g., setting process.env, to ensure that each side fires when intended. There are examples of detecting that a fetch fired (even if it's non-local) in the tests. I believe Yanwen has a PR or draft open that contains such an example. You can search through closed PRs by setting the search syntax. It makes sense when you look at it.

packages/backend/src/functions/httpTrigger.ts
HieuNgoGH: Hey @paulErdos 

> Be sure to include the other two, gosqas.org and dev.gosqas.org. Could then write: const location = frontendUrl.includes('red') || frontend.includes('dev') ? 'production' : 'staging';
> 
> and use that in your first if, and make the second one an else. Better yet, return at the end of the first if, and free the second from the conditional.

Does that mean I should also fetch gosqas.org and dev.gosqas.org or just include them in the frontendUrl check?

paulErdos: Yeah, we wanna fetch em.

packages/backend/src/functions/httpTrigger.ts
```diff
@@ -1707,12 +1709,60 @@ export async function addEntryHandler(request: HttpRequest, context: InvocationC
...
+        await livenessCheckEmailer('Production')
+        }
+
+    // On prod and checking if dev is down then send email staing dev is down
+    if ((frontendUrl.includes('blue') || frontendUrl.includes('https://gosqas.org/')) && stageResponse.status !=200) {
+        await livenessCheckEmailer('Staging')
+        }
+}
+
+export async function livenessCheckEmailer (str: string, context?: InvocationContext) {
```
HieuNgoGH: Will change variable str to serverName to be clearer. Will put in draft until then.

packages/backend/src/functions/httpTrigger.ts
```diff
@@ -1728,12 +1709,60 @@ export async function addEntryHandler(request: HttpRequest, context: InvocationC
...
+
+    // On prod and checking if dev is down then send email staing dev is down
+    if ((frontendUrl.includes('blue') || frontendUrl.includes('https://gosqas.org/')) && stageResponse.status !=200) {
+        await livenessCheckEmailer('Staging')
+        }
+}
+
+export async function livenessCheckEmailer (server: string, context?: InvocationContext) {
+
+    const emails = process.env['LIVENESS_CHECK_EMAIL_RECIPIENTS']
```
paulErdos, line 1807: FYI on this one. At time of writing, we have:

```
   const emails = process.env['LIVENESS_CHECK_EMAIL_RECIPIENTS']
   try {
        for (const email of emails) {
            // do stuff
        }
   }
```

But that results in this:

Image

paulErdos (approved):

packages/backend/src/functions/httpTrigger.ts
paulErdos: fyi @HieuNgoGH

PR #1212, opened 2026-08-28
[Bug Fix] CSV Erroneously Treats First Child Key as Public Key 

### LINKED ISSUE #480: [BUG] CSV file of children erroneously treats the first child key as the reporting key whether it is or not....
This came from @paulErdos 's question in the Pull Request:
[pull request](https://github.com/gosqasorg/asset-provenance-tracking/pull/478)

The image below demonstrates the problem (created from a record which does NOT have a reporting key.)

Image

I suspect @deneille and @paulErdos will know how to fix this very quickly, so I am assigning to them.

HieuNgoGH:
Modified code to have public key set to empty string instead of first child key when there is no public key requested by user.

With public key:
with_public_key

Without public key:
no_public_key

### DIFF
--- a/packages/frontend/components/csvFile.vue
+++ b/packages/frontend/components/csvFile.vue
@@ -24,7 +24,7 @@ export default {
                 const childrenKeys = await getChildrenKeys(this.recordKey);
 
                 const publicRecord = provenance?.[0]?.record;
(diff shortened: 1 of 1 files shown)


paulErdos:
Success. 

Screenshot 2026-09-14 at 09 17 03

paulErdos (approved):

PR #1213, opened 2026-08-29
Subscribing to Group Notifies You of Children

### LINKED ISSUE #383: [FEATURE REQUEST] Subscribing to a group notifiies you of all the descendants.
Suppose P is a a group of 100 tourniquets, C0, C1, ... C100, with a reporting key R,

If you subscribe to the REPORTING KEY R (not the parent P), then the subscriber should receives notifications of ANY descendant as if they had subscribed to all of the descendants.

agilpatr:
Since children keys do not store parent keys, the way I found to allow group subscribers to get email notifications for all children updates was by subscribing group emails to all descendants. I've updated our subscribe/unsubscribe functions to loop through all children of the record we're looking at and subscribe/unsubscribe them. I've also updated adding group/children keys to subscribe group users to the new key.
sub:unsub_problem

For recall/send to all children, for now I've made it so that only the group is notified (since the alternative is emailing every child, which risks flooding people's emails over one record created [which would both take a lot of time and if we're limited on number of email sends potentially use them up]). In future it would likely be worth updating this to send a specific recalled/sent to all children email, and to notify subscribers only subscribed to one of the children.

There are a few cons with this implementation:
- To subscribe newly added children I needed to create an endpoint directly for subscribing, which bypasses our email verification
- If you add a new child record and it fails to subscribe, then the group will never be notified about that child's updates


packages/backend/src/functions/emailNotificationUtils.ts
```diff
@@ -154,63 +154,78 @@ export async function subscribeToNotifications(containerClient: ContainerClient,
...
-    let [emailSet, emailIDSet] = await getExisitingEmails(exists, blobClient);
+    // Loop through record/children keys and subscribe to all of them
+    while (keysToCheck.length != 0) {
+        // Add any children of the current key to the list to subscribe to
+        let key = keysToCheck[0];
+        let uniqueChildKeys = await getChildKeys(key, deviceKey);
+        keysToCheck = keysToCheck.concat(uniqueChildKeys);
+
+        // Setup the blobClient and get emails subscribed to the record
+        let [blobName, blobClient] = await setupBlobClient(containerClient, calculateDeviceID, key);
```
paulErdos, line 173: It is unclear why we're passing around the function calculateDeviceID. It can be imported and used where needed.

packages/backend/src/functions/emailNotificationUtils.ts
```diff
@@ -154,63 +154,78 @@ export async function subscribeToNotifications(containerClient: ContainerClient,
...
+        const exists = await blobClient.exists();
```
paulErdos, line 174: The check for existence should probably be added to setupBlobClient. It should not be a parameter to getExistingEmails, which without exists, is a no-op that hands back empty lists.

packages/backend/src/functions/emailNotificationUtils.ts
```diff
@@ -154,63 +154,78 @@ export async function subscribeToNotifications(containerClient: ContainerClient,
          - https://learn.microsoft.com/en-us/javascript/api/%40azure/storage-blob/blockblobuploadoptions?view=azure-node-latest
     */
 
-    // Setup the blobClient
-    let [blobName, blobClient] = await setupBlobClient(containerClient, calculateDeviceID, deviceKey);
-    const exists = await blobClient.exists();
+    let keysToCheck = [deviceKey];
 
     // Confirm the email exists
     const normalized = (email ?? "").trim().toLowerCase();
```
paulErdos, line 160: This should be renamed to be more specific, e.g. "normalizedEmail" or "normalizedEmailAddress".

packages/backend/src/functions/emailNotificationUtils.ts
```diff
@@ -154,63 +154,78 @@ export async function subscribeToNotifications(containerClient: ContainerClient,
...
+            {
+            name: "AES-CBC",
+            length: 256
+            },
+            true,
+            ['encrypt', 'decrypt']
+        );
+
+        const buffer = await crypto.subtle.exportKey("raw", uniqueString);
+        const uniqueEmailString = base58encode(new Uint8Array(buffer));
```
paulErdos, line 198: This should be renamed emailID. Discussion of it being unique is better for a comment.

packages/backend/src/functions/emailNotificationUtils.ts
```diff
@@ -154,63 +154,78 @@ export async function subscribeToNotifications(containerClient: ContainerClient,
...
+        emailIDSet.add(uniqueEmailString)
+
+        try {
+            // Update our stored emails to include the new email/id
+            uploadBlob(containerClient, blobName, emailSet, emailIDSet, tags);
```
paulErdos, line 203: If exists is false, this may likely always land in the catch block. It's probably worth thinking about how checking the blob client for whether the underlying blob exists. It may be best to throw an error and return 400 with an informative error message.

packages/backend/src/functions/emailNotificationUtils.ts
```diff
@@ -154,63 +154,78 @@ export async function subscribeToNotifications(containerClient: ContainerClient,
...
 export async function unsubscribeFromNotifications(containerClient: ContainerClient, calculateDeviceID: (key: string | Uint8Array) => Promise<string>, deviceKey: string, emailID: string, tags: string[] = []) {
-    // Setup the blobClient
-    let [blobName, blobClient] = await setupBlobClient(containerClient, calculateDeviceID, deviceKey);
-    const exists = await blobClient.exists();
+    // Get the email that corresponds to the given email id
+    let keysToCheck = [deviceKey];
 
-    let [emailSet, emailIDSet] = await getExisitingEmails(exists, blobClient);
+    const [blobName, blobClient] = await setupBlobClient(containerClient, calculateDeviceID, deviceKey);
+    const exists = await blobClient.exists();
```
paulErdos, line 225: Just noting instances of relevance: the above with regard to exist applies here.

packages/backend/src/functions/emailNotificationUtils.ts
```diff
@@ -219,27 +234,50 @@ export async function unsubscribeFromNotifications(containerClient: ContainerCli
...
+    // Loop through record/children keys and unsubscribe from all of them
+    while (keysToCheck.length != 0) {
+        // Add any children of the current key to the list to unsubscribe from
+        let key = keysToCheck[0];
+        let uniqueChildKeys = await getChildKeys(key, deviceKey);
+        keysToCheck = keysToCheck.concat(uniqueChildKeys);
+
+        // Setup the blobClient and get emails subscribed to the record
+        let [blobName, blobClient] = await setupBlobClient(containerClient, calculateDeviceID, key);
+        const exists = await blobClient.exists();
```
paulErdos, line 246: Just noting instances of relevance: the above with regard to exist applies here.

packages/backend/src/functions/emailNotificationUtils.ts
```diff
@@ -154,63 +154,78 @@ export async function subscribeToNotifications(containerClient: ContainerClient,
...
+        let [emailSet, emailIDSet] = await getExisitingEmails(exists, blobClient);
+
+        // Add the specified email to the set
+        const sizeBeforeAdding = emailSet.size;
+        emailSet.add(normalized);
+
+        // If email is already stored move on to the next key to check
+        if (exists && emailSet.size === sizeBeforeAdding) {
```
paulErdos, line 182: The goal is to see if the email is already in the set. We don't want to store the size, add, and see if the size is the same to determine set membership. We can directly check if the Set.has(something).

packages/backend/src/functions/httpTrigger.ts
```diff
@@ -1562,6 +1554,45 @@ export async function createGroupHandler(request: HttpRequest, context: Invocati
...
+            jsonBody: {error: "Error: group and child keys required"},
+            status: 400
+        }
+    }
+
+    await containerClient.createIfNotExists();
+
+    // 1. Get emails subscribed to the groupKey
+    let [blobName, blobClient] = await setupBlobClient(containerClient, calculateDeviceID, groupKey);
+    const exists = await blobClient.exists();
```
paulErdos, line 1575: This also is consistent with the above if not exist, return 400 + informative error message.

packages/backend/src/functions/httpTrigger.ts
```diff
@@ -1562,6 +1554,45 @@ export async function createGroupHandler(request: HttpRequest, context: Invocati
...
+    let [emailSet, emailIDSet] = await getExisitingEmails(exists, blobClient);
```
paulErdos, line 1576: Let's have getExistingEmails(groupKey, containerClient) and have getExistingEmails handle the setupBlobClient work.

packages/backend/src/functions/httpTrigger.ts
```diff
@@ -1562,6 +1554,45 @@ export async function createGroupHandler(request: HttpRequest, context: Invocati
...
+
+    // 2. Subscribe all emails from the groupKey to the childKey
+    for (let email of emailSet) {
```
paulErdos, line 1579: Good, correct.

paulErdos (changes requested):

PR #1215, opened 2026-08-31
Updated Offline Queue Stash

### LINKED ISSUE #1185: Update our Queue/Failed Stashes
Currently our queue stash is laid out like this:
   - gosqas-offline-stash-1 = [key, formData]
   - gosqas-offline-stash-2 = [key, formData]
   - ...

And our failed stash is laid out like this:
   - gdt-stash-failed = [[key, formData], [key2, formData2], ...]

What we want to do is give these two stashes the same layout format, so that they're easier to work with and understand later. The new layout we want to use is as follows:
   - gdt-stash-queued = [{“key”: key, “data”: formData}, {“key”: key2, “data”: formData2}, ...]
   - gdt-stash-failed = [{“key”: key, “data”: formData}, {“key”: key2, “data”: formData2}, ...]

We're using a list of dictionaries instead of a list of lists so that when grabbing values from the stash it is clearer to the devs what we are grabbing (queueStash[0][1] vs queueStash[0][“key”]).

### LINKED ISSUE #1191: Update stashOfflineRequest/removeOfflineRequest to Work for All Stashes
Our stashOfflineRequest and removeOfflineRequest functions currently work for the syncing, failed, and fulfilled stashes (since our old queue stash was formatted too differently to modify here). Now that our failed and queue stashes are the same, we'll want to update these functions to (1) use the new stash format shown below and (2) handle queue requests.

Old failed stash format:
- gdt-stash-failed = [[key, formData], [key2, formData2], ...]

New failed/queue stash format:
- gdt-stash-failed = [{“key”: key, “data”: formData}, {“key”: key2, “data”: formData2}, ...]
- gdt-stash-queued = [{“key”: key, “data”: formData}, {“key”: key2, “data”: formData2}, ...]

We want to remove the syncing stash from these functions (since we're no longer using it).

We'll also want to add some tests to azureFuncs.spec.ts to make sure these functions are working as intended (see [our old tests](https://github.com/gosqasorg/asset-provenance-tracking/issues/1189), as there are likely many of these that could be used to test this function).

agilpatr:
Implemented the queue stash (gdt-stash-queued) and updated the failed stash layout in stashOfflineRequest and removeOfflineRequest. I updated our offline edits page to be able to display our new stash layout format, and wrote some tests to make sure the new layouts are working as intended.

The new layout is this (I used the record instead of formData since we can only store strings in localStorage:
- gdt-stash-queue = [{“key”: key, “data”: record2}, {“key”: key2, “data”: record2}, ...]
- gdt-stash-failed = [{“key”: key, “data”: record2}, {“key”: key2, “data”: record2}, ...]

Note: Since there was a lot of overlap between issues #1185 and #1191 I've gone ahead and solved both in this PR.

### DIFF
--- a/packages/frontend/pages/offline-edits.vue
+++ b/packages/frontend/pages/offline-edits.vue
@@ -59,7 +59,7 @@ while offline.
         </div>
     </div>
     <!----------------Queued Keys Banner-------------------->
-    <div v-for="(key, index) in offlineKeys">
+    <div v-for="(key, index) in queuedKeys">
         <div class="key-box">
             <p style="font-size: 17px;">{{ key }}</p>
             <div style="background-color: #91bdf5; border-radius: 20px; width: 105px; color: black; text-align: center; height:40px; display:flex; justify-content: center; align-items: center;">Queued</div>
@@ -99,7 +99,7 @@ data() {
         dismissAllEditsPopUp: false,
         dismissPublishedEditsPopUp: false,
         dismissSingleEditPopUp: false,
-        offlineKeys: [] as string[],
+        queuedKeys: [] as string[],
         fulfilledKeys: [] as string[],
         failedKeys: [] as string[],
 	}
@@ -108,7 +108,7 @@ data() {
 async mounted() {
     try {
         this.getFailedKeys();
-        this.getOfflineKeys();
+        this.getQueuedKeys();
         this.getFulfilledKeys();
         this.clearOneEdit();
     } catch (error) {
@@ -121,14 +121,17 @@ async mounted() {
 },
 
 methods: {
-    getOfflineKeys() {
-        // Get all keys that were successfully stashed while offline
-        let stash_counter = parseInt(localStorage.getItem('stash_counter') || "0");
-        for (stash_counter; stash_counter > 0; stash_counter--) {
-            let stashedRequest = JSON.parse(localStorage.getItem('gosqas-offline-stash-' + stash_counter) || '{}')
-            let fullUrl = stashedRequest[0][1]
-            let record = fullUrl.split("/")[fullUrl.split("/").length - 1]
-            this.offlineKeys.push(record)   
+    getQueuedKeys() {
+        let queued = localStorage.getItem("gdt-stash-failed") || '{}';
+        if (queued == '[{}]' || queued == '{}') {
+            return
+        }
+
+        for (const request of JSON.parse(queued)) {
+            if (JSON.stringify(request) !== "{}") {
+                let key = request["key"];
+                this.queuedKeys.push(key);
+            }
         }
     },
     getFulfilledKeys() {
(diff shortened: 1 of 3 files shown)


packages/frontend/pages/offline-edits.vue
```diff
@@ -121,14 +121,17 @@ async mounted() {
...
-        // Get all keys that were successfully stashed while offline
-        let stash_counter = parseInt(localStorage.getItem('stash_counter') || "0");
-        for (stash_counter; stash_counter > 0; stash_counter--) {
-            let stashedRequest = JSON.parse(localStorage.getItem('gosqas-offline-stash-' + stash_counter) || '{}')
-            let fullUrl = stashedRequest[0][1]
-            let record = fullUrl.split("/")[fullUrl.split("/").length - 1]
-            this.offlineKeys.push(record)   
+    getQueuedKeys() {
+        let queued = localStorage.getItem("gdt-stash-queued") || '{}';
+        if (queued == '[{}]' || queued == '{}') {
```
paulErdos: This is handled by the loop (either hitting {} or nothing, respectively) and the check in the loop (which checks, for the looping over [{}] case, whether request is {}).

packages/frontend/pages/offline-edits.vue
```diff
@@ -121,14 +121,17 @@ async mounted() {
...
+            return
+        }
+
+        for (const request of JSON.parse(queued)) {
+            if (JSON.stringify(request) !== "{}") {
```
paulErdos: Recommendation: if == {}: continue.

packages/frontend/pages/offline-edits.vue
paulErdos: This is the same as getQueuedKeys. It appears that getFulfilledKeys can follow this same workflow. Consider making this a utility function, "getKeysFromStash(stashName)", for example, and calling it with the stash name as the parameter. Then these three functions can simply call that utility function.

packages/frontend/services/azureFuncs.ts
```diff
@@ -303,43 +303,43 @@ export function stashOfflineRequest(currentKey: string, stashName: string, reque
...
+        if (JSON.stringify(existingRequests) == "[]" || JSON.stringify(existingRequests) == '["[]"]') {
             return;
         }
 
-        if (stashName.includes("failed")) {
-            // Remove request from failed stash
+        if (stashName.includes("fulfilled")) {
+            // Remove key from the fulfilled stash
+            const index = existingRequests.indexOf(currentKey);
+            if (index > -1) {
```
paulErdos:
```
>= 0
```

paulErdos:
Note to self: reviewed all but test.

paulErdos (changes requested):

paulErdos (approved):

PR #1222, opened 2026-09-05
Implement confirmRequestFulfilled Function

### LINKED ISSUE #1187: Implement confirmRequestFulfilled Function
Goal: two cases:
1. Record or group creation --> Ensure creation success
2. Record history entry addition --> Ensure creation success

We want to create a function that will take information about a record (deviceKey and record) and attempt to fetch it. It should then confirm that the most recent record entry matches the record from our parameters (we want this to be able to check both new records/groups and new record entries were successfully posted). It should return a boolean value.

We'll also want to write some simple tests for this function to confirm it is working as intended. Add them to the azureFuncs.spec.ts file.

HieuNgoGH:
- Implemented confirmRequestFulfilled function in azureFuncs
- Implemented tests in azureFuncs.spect

### DIFF
--- a/packages/frontend/services/azureFuncs.ts
+++ b/packages/frontend/services/azureFuncs.ts
@@ -312,6 +312,26 @@ export function removeOfflineRequest(currentKey: string, stashName: string) {
     }
 }
 
+export async function confirmRequestFulfilled(recordKey: string, record?: any): Promise<boolean> {
+    try {
+        let response = await getProvenance(recordKey)
+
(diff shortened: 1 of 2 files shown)


agilpatr (approved):
Hey Hieu, everything looks good here! Just a note, I noticed this PR was tied to the Spec page issue, rather than the confirmRequestFulfilled issue. I've switched this PR to instead close the confirmRequestFulfilled issue and left the Spec page one open, though reach out to me if this was intentional.

PR #1223, opened 2026-09-11
Added Offline Tolerance to getProvenance

### LINKED ISSUE #1151: Add fault tolerance to getProvenance for compatibility with Offline Mode
Note: an open question that underlies this ticket is how new record history elements are added to the provenance json as a whole. Inspecting the post, just the new record is sent to the backend. Inspecting the frontend and backend code (azurefuncs, httptrigger), we find that where these are joined is non-obvious, and is where we will pick up next time. 

Update: by manual inspection, we have found the following:
* The top-level record (the device itself) has its own provenance json, and record history updates themselves are not contained inside this json structure, but rather are returned from the backend alongside, e.g., [update3, update2, update1, device_json]. 
* This mirrors what we see in the backend code. Note how getProvenance goes into the *folder* path "prov/<some_device_id>, and inside that folder reside the device json, and the collection of record history elements, all timestamped so they can be sorted by the page. 

----

One of our interests is to make the history page know nothing about offline mode or connectivity. This will allow us to use the existing history page, reducing labor demand in regards to both creating, and importantly also, testing + maintaining the separate page long term. 

The idea is this:
Since we can detect connectivity by the return code being undefined, we can we adapt services/AzureFuncs.ts --> getProvenance to execute the fetch request as normal, and if offline, to then query the stash. If the item is not in the stash, then we act as normal for a not found key. However, if it is, then we can hand the caller, importantly, the history page (pages/history/[deviceKey].vue), the json in exactly the same way as though we were online. 

We'll want to write a function that takes a device key and queries all stashes for any json matching that device key. It should hand back an array of record entries and the json for the device itself (note: we're expecting that the device/entries will have timestamps, this is not implemented yet but will be in a separate ticket).

The principal implication of this is a double-edged sword that is dull on one side. On the good side, whatever side of a sword that is, everything that calls getProv is immediately compatible with offline mode. And this is important, because (at present) the history page calls getProv in three places, and includes the very nice side benefit of making everything else that relies on getProv compatible with one function edit. 

The downside is that we then must check everything in the frontend that calls getProv. At present there are 18 calls, so 15 extra. 

If we make sure we are careful and thorough with testing the new getProv, we can, rather than testing everything, put a few tests into the test suite, and for due diligence, manually test everything once just to be sure. Deep type annotation (ie, not using "any"), or actually useful, zod validation to ensure that getProvenance always returns the same types (ie, not just array, but array of the right things including eg checking for the timestamp) is another approach to handling testing, and with this, there would be no need to write eg 18 tests. 

As a side note, one major time saving implication of having a test suite is that if everything in the frontend were tested (we're on our way) then we could introduce a possibly-breaking change, run the suite, and know whether or not anything breaks in one action. 

The above (and screenshot below) is a simplification. What we really want to do is write a separate function that queries all of the stashes, and have getprov call this function similar to the screenshot below. (thanks for noting this subtlety @agilpatr!)

Example:
Image

agilpatr:
I've updated getProvenance to call a new function getProvenanceOffline if it fails to fetch the records while offline. This function checks all three stashes (queued, failed, fulfilled) for keys matching the given key and compiles their stored data into a list based on timestamp to display. If no matches are found it returns an error to display as normal. I've also updated postProvenance to stash the record whenever it fails to post offline.

To make getProvenanceOffline work I updated all three stashes to have the same format below:
[{"key": currentKey, "data": request, "timestamp": timestamp}]
We create a timestamp when we stash the request, which allows us to accurately sort them on the history page.

With these two changes you can now open the PWA, go offline, and create records + add record entries to them on the history page. I wrote some tests for the new features and manually tested every instance of getProvenance both online and offline.

### DIFF
--- a/packages/frontend/pages/offline-edits.vue
+++ b/packages/frontend/pages/offline-edits.vue
@@ -59,7 +59,7 @@ while offline.
         </div>
     </div>
     <!----------------Queued Keys Banner-------------------->
-    <div v-for="(key, index) in offlineKeys">
+    <div v-for="(key, index) in queuedKeys">
         <div class="key-box">
             <p style="font-size: 17px;">{{ key }}</p>
             <div style="background-color: #91bdf5; border-radius: 20px; width: 105px; color: black; text-align: center; height:40px; display:flex; justify-content: center; align-items: center;">Queued</div>
@@ -99,7 +99,7 @@ data() {
         dismissAllEditsPopUp: false,
         dismissPublishedEditsPopUp: false,
         dismissSingleEditPopUp: false,
-        offlineKeys: [] as string[],
+        queuedKeys: [] as string[],
         fulfilledKeys: [] as string[],
         failedKeys: [] as string[],
 	}
@@ -108,7 +108,7 @@ data() {
 async mounted() {
     try {
         this.getFailedKeys();
-        this.getOfflineKeys();
+        this.getQueuedKeys();
         this.getFulfilledKeys();
         this.clearOneEdit();
     } catch (error) {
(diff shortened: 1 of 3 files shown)


agilpatr:
**Note:** I'm moving this back to draft until the offline worker is merged in (since this PR modifies the fulfilled branch, so I'll have to make some small updates to the worker).

PR #1228, opened 2026-09-15
Alison queue first 1188

### LINKED ISSUE #1188: Implement getFirstQueueItem/removeFirstQueueItem Functions
We need two new functions, one that non-destructively retrieves the first value from the queue stash and another that destructively removes it.
- getFirstQueueItem should take nothing and return the first item from the queue stash (without removing it from the stash)
- removeFirstQueueItem should take nothing and remove the request from the front end (FIFO) of the queue stash

We'll want to write some basic tests for these functions as well in the azureFuncs.spec.ts file.

paulErdos:


paulErdos:
Reviewed in office hours

PR #1229, opened 2026-09-16
1217 to speedup return times for usage data create data cache object

### LINKED ISSUE #1217: To Speedup Return Times for Usage Data, Create Data Cache Object
One of our design goals is to create gosqas such that any user who wishes to fork our repository and spin up their own instance of GDT has a minimal workload insofar as setting up the cloud backend. 

We wish to support this, and in doing so, we can also speed up return times for data access. 

We can do this by defining a class, the member functions of which run the slow data access functions, and then store the data in member variables. Time triggers can be set to call the update functions on an hourly basis. Then, when frontend callers need data, they can call a getter function which hands back the cached data. 

Presently, e.g., getStatistics[1], setStatisticsTotals[2], and getBrowserStats need this.

SetStatisticsTotals stores what in finds in the cloud, in particular, in blob storage. This is an excellent first start, however, we wish to use the cloud as little as possible, and its use here should be eliminated. SetStatisticsTotals should  store data in the instantiated data object. 

GetStatistics then refers to data stored in blob store by setStatisticsTotals. It also goes and grabs a great deal of other data. The data-gathering (ie, backend querying) operations must be moved into a (time trigger updated) function in the data class, and getStatistics can then be simplified to:
1. Grab the data for which it typically queries the backend (separate from the blob storage query to get the data grabbed by SetStatisticsTotals)
2. Grab the data that SetStatisticsTotals added to the data class, then
3. Return these data. 

GetBrowserStats must also be updated in this way: its data acquisition operations should be moved to a member function in the data class, and getBrowserStats should then reference this data class. 

This data class will, as the project progresses, expand to be our centralized usage data reference object. 

GetStatistics 
https://github.com/gosqasorg/asset-provenance-tracking/blob/main/packages/backend/src/functions/httpTrigger.ts#L493

SetStatisticsTotals
https://github.com/gosqasorg/asset-provenance-tracking/blob/main/packages/backend/src/functions/httpTrigger.ts#L613

GetBrowserStats:
https://github.com/gosqasorg/asset-provenance-tracking/blob/main/packages/backend/src/functions/getStats.ts#L56

mishwilso:
Added:
- statCache.ts which has our cache class. It holds totals, queryStats, and browserStats, and each has an update function that refreshes it.
- refreshStats, which has a timer trigger that calls the update functions once an hour and once on startup.

Updated getStatistics and getBrowserStats to now read from statsCache instead of querying directly.
Removed the setStatisticsTotal function, and moved getStats.ts logic to statCache.ts


packages/backend/src/utils/statsCache.ts
```diff
@@ -0,0 +1,224 @@
...
+                    "Content-Type": "application/json"
+                },
+                body: JSON.stringify({ query })
+            }
+        );
+
+        const data = await result.json();
+        return data.tables[0].rows
+    } catch(error) {
+        context.log(`Leaving runQuery: error occurred: ${error}`)
```
paulErdos: Use context.error here.

packages/backend/src/utils/refreshStats.ts
```diff
@@ -0,0 +1,32 @@
...
+        }
+    }
+    
+    context.log('Usage stats cache refreshed');
+}
+
+// Timer trigger to refresh usage stats every hour, and on startup.
+app.timer('refreshUsageStats', {
+    schedule: '0 0 * * * *',
+    // schedule: '*/10 * * * * *', // for testing runs every 10 secs
```
paulErdos: Testing-related comments are best to drop before PR

packages/backend/src/utils/refreshStats.ts
```diff
@@ -0,0 +1,32 @@
+import { app, InvocationContext, Timer } from '@azure/functions';
+import { usageStatsCache } from './statsCache.js';
+import { containerClient } from '../functions/httpTrigger.js';
+
+
+export async function usageRefresh (myTimer: Timer, context: InvocationContext, ): Promise<void> {
```
paulErdos: We can probably drop this myTimer argument

packages/backend/src/utils/refreshStats.ts
```diff
@@ -0,0 +1,32 @@
...
+    context.log('Refreshing usage stats cache');
+    // allSettled so one failing update can't cause the invocation to complete
+    // while the others are still running in the background. 
+    const results = await Promise.allSettled([
```
paulErdos: Try doing concurrency here by calling outside of allSettled, and not awaiting any.

packages/backend/src/utils/refreshStats.ts
paulErdos: Go ahead and move this into the statsCache class

packages/backend/src/utils/refreshStats.ts
```diff
@@ -0,0 +1,32 @@
...
+        if (result.status === 'rejected') {
+            context.log('Error refreshing usage stats cache', result.reason);
+        }
+    }
+    
+    context.log('Usage stats cache refreshed');
+}
+
+// Timer trigger to refresh usage stats every hour, and on startup.
+app.timer('refreshUsageStats', {
```
paulErdos: Go ahead and move this into statsCache.ts

paulErdos:
Mish is taking a look into what tests exists for the relocated functions and will note what testing gaps if any exist.

paulErdos (changes requested):

